PR #261, opened 2024-08-19
fix(EG-616): file upload

simondib:
Removes upload timeout preventing large files from uploading. 

Simplifies and improve display logic for remove button, upload button and dropdown.


zigtan (approved):
Thanks @simondib .

PR #262, opened 2024-08-20
fix(EG-653): update config to inc 'seqera-api-base-url' setting & root 'build-and-deploy' convenience script

zigtan:
This PR updates the `easy-genomics.example.yaml` and README.md files to inc the optional `'seqera-api-base-url' setting in case the AMD training wishes to configure Easy Genomics with a self-hosted NextFlow Tower instance.

It also adds a sanity logic to ensure that a trailing slash for `'seqera-api-base-url' setting is removed.

It also adds the root project convenience script `'build-and-deploy'` to allow both the FE and BE to be deployed in one command once the initial BE and FE have been deployed.

### DIFF
--- a/packages/back-end/src/main.ts
+++ b/packages/back-end/src/main.ts
@@ -155,7 +155,7 @@ if (process.env.CI_CD === 'true') {
         systemAdminPassword: systemAdminPassword,
         testUserEmail: testUserEmail,
         testUserPassword: testUserPassword,
-        seqeraApiBaseUrl: seqeraApiBaseUrl,
+        seqeraApiBaseUrl: seqeraApiBaseUrl.replace(/\/+$/, ''), // Remove trailing slashes
       });
     }
   });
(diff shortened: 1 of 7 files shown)


packages/back-end/src/main.ts
```diff
@@ -155,7 +155,7 @@ if (process.env.CI_CD === 'true') {
         systemAdminPassword: systemAdminPassword,
         testUserEmail: testUserEmail,
         testUserPassword: testUserPassword,
-        seqeraApiBaseUrl: seqeraApiBaseUrl,
+        seqeraApiBaseUrl: seqeraApiBaseUrl.replace(/\/+$/, ''), // Remove trailing slashes
```
PurcyTwoBulls, line 158: Could you just use trim here?

zigtan: @PurcyTwoBulls , sorry no trim is just for removing whitespaces. This preventive fix is to remove any potential trailing slashes as we already insert a `'/'` when we prepare the NextFlow API URLs.

Additionally, we explicitly document that the values in the `easy-genomics.yaml` file be enclosed in quotes to prevent whitespaces.

PurcyTwoBulls (approved):

PR #265, opened 2024-08-22
feat(EG-662): improve eslint + prettier

simondib:
This PR aims to improve the existing eslint and prettier implementation in order to promote best practice code patterns and consistent code formatting.

### **Configuration:**

Prettier and ESLint configuration is via the Projen config file.  After running `pnpm projen` their config, rules and ignore lists will be generated. 

**Prettier** uses a single rule file (prettierrc.json) that applies fomatting to the entire codebase for `js,ts,vue,scss,json,md,html,mdx` file types.

**ESLint** shares a global rule list for all three packages, but each package can further have their rules tweaked as required.

### **Execution:**

**Prettier** is run via `pnpm prettier` from the root of the project, preconfigured to scan and fix the previously mentioned filetypes. It will also run as a pre-commit githook for any staged files prior to committing to Git.

**ESLint** can be run per package, e.g. `packages/front-end/pnpm lint` or from the root of the project as `pnpm lint-all` to lint all three packages, i.e. `front-end`, `back-end` and `shared-lib`.

### **Other related changes:**

- fixed package aliases in Projen for `@FE`, `@BE`, and `@SharedLib` that can now be used in-code correct. Also refactored existing `~/` path aliases for some imports to allow ESLint to correctly resolve these paths when from the root.
- bumped ES2019 -> ES2022 version globally in tsconfig to leverage more modern ECMAScript
- linted and formatted codebase against the ESLint and Prettier rules


zigtan (approved):
Great work! Thanks heaps @simondib !

PurcyTwoBulls (approved):

PR #266, opened 2024-08-22
fix(EG-664): update lab role in UI  on change

simondib:
Fixes the lab role not updating in the UI when changed between "Lab Manager" and "Lab Technician".


zigtan (approved):
Thanks heaps @simondib .

PR #267, opened 2024-08-22
fix: use case insensitive sorting for name of users, labs and orgs

PurcyTwoBulls:
Overrides the sort function to apply a case insensitive sort (with a case sensitive sort for consistent results if they match).

This is applied to all sortable text fields in tables.


packages/front-end/src/app/utils/sort-utils.ts
```diff
@@ -0,0 +1,19 @@
...
+      return a > b ? -1 : 1;
+    }
+    return 0;
+  }
+  if (direction === 'asc') {
+    return a.toLowerCase() < b.toLowerCase() ? -1 : 1;
+  } else {
+    return a.toLowerCase() > b.toLowerCase() ? -1 : 1;
+  }
+}
```
zigtan, line 19: Thanks @PurcyTwoBulls for working on this fix. This function needs some improvement.

Organization Names, Laboratory Names, and also User Emails are enforced to be unique within their respective DynamoDB tables with the use of the supporting `unique-reference-table` DynamoDB table. 

Have a look through this records to see the unique references - you will see that Laboratory Names are unique within their parent Organization.

So, if there is an existing Laboratory named `'ABC'` the backend logic will never allow a new Laboratory named `'abc'` to be created - as the lower-case uniqueness record already exists in the `unique-reference-table`.

As a result, this function can be simplified to:

```
export function caseInsensitiveSortFn(a: string, b: string, direction: 'asc' | 'desc'): number {
  const orderAB = a.toLowerCase() < b.toLowerCase() ? -1 : 1;
  return direction === 'asc' ? orderAB : orderAB * -1; // If desc, invert orderAB
}
```

Also we should not ever need to use the `any` type, as we are using Typescript instead of Javascript so we should be explicitly defining the types.

zigtan (changes requested):
Thanks @PurcyTwoBulls . Just need to update the FE sort-utils `caseInsensitiveSortFn()` logic.

packages/front-end/src/app/utils/sort-utils.ts
PurcyTwoBulls: Thanks @zigtan 

For users, emails have uniqueness but the users first and last name and combinations do not. If we want to make the decision that we are not tackling that edge case then we don't need to, but given it is user entry it is probable edge case.

I have updated the function signature.

zigtan: @PurcyTwoBulls , a User's names (`FirstName`, `LastName`) do not have to be unique in the system. 

For all User's their `FirstName`, `LastName` are just their personal details. 

Internally we reference all User's by their primary UUID.

The User's Email is used as a Global Secondary Index (GSI) for efficient lookups in DynamoDB.

The rule to ensure each User's Email is unique a logical security design and also support the multi-tenant architecture where a User can invited to one or more Organizations (using the same email).

PurcyTwoBulls: I understand that, but what is being used here for displaying user is their display name or their email address, so we are using First and Last name in these sorted fields.

zigtan: Thanks @PurcyTwoBulls for highlighting the First Name and Last Name sorting issue you are trying to tackle. 

I think this is a deeper problem seeming from our UI design.

We will need to circle back to improve the UI design to separately list Users' `Email`, and `Full Name` in separate columns and have each be sortable. The current design of showing a single column for `Email` or `Full Name` complicates the sorting behaviour and solution.

zigtan (approved):

PR #272, opened 2024-08-28
fix(EG-668): lint package folders separately to fix context issue

simondib:
Running `lint-all` from root was causing an issue between ESLint/tsconfig establishing the right folder path/context, throwing an error and preventing the ability to commit. 

The simplest solution was to separate the lint tasks out separately and run sequentially as part of the existing `lint-staged` script. The Prettier step has been moved to run before ESLint also as best practice.


zigtan (approved):
Thanks heaps @simondib !

PR #273, opened 2024-08-29
feat: new api for testing nextflow tower credentials

PurcyTwoBulls:
Takes a WorkflowId and an AccessToken before testing a call to compute-envs for nextflow tower.

200 response with status of `Success`
400 response if we couldn't make a successful call to nextflow tower, or if Invalid request for missing inputs (can't give better responses unless we add custom error types when throwing errors)

I moved this into the nextflow tower api path because it felt more approprate and also because it would get the correct env variables to get the nextflow api endpoint.


zigtan (approved):
Thanks @PurcyTwoBulls .

PR #274, opened 2024-08-29
fix: apply apache license with correct copyright name and year

PurcyTwoBulls:
Extends the Projen License class to apply the Apache-2.0 and override the year and owner placeholders.


zigtan (approved):
Thanks @PurcyTwoBulls , nice work.

PR #277, opened 2024-09-01
feat(EG-641): change default labs tab to Runs

simondib:
Changes Pipelines tab to Runs tab as the first/default tab when viewing the Lab.


zigtan (approved):
Thanks @simondib .

PR #278, opened 2024-09-02
feat(EG-657): force input of lab workspace id and personal access token

simondib:
This PR enforces a workspace id and personal access token (PAT) once a Lab is created in order to access the full functionality of a Lab. 

If a workspace id or PAT are not detected after accessing a Lab a modal will inform the user to add these details. Display logic hides the existing Lab tabs (Pipelines, Runs, Lab Users) until these conditions are satisfied.

In order to validate these details when entered a check is made using the id and PAT values against `/nf-tower/connection/request-connection-test` endpoint. If successful, the Lab tabs are made available and full Lab functionality is granted to the user.


zigtan (approved):
Thanks @simondib , this is great!

PR #279, opened 2024-09-02
chore: use correct headers for error responses

PurcyTwoBulls:
Uses the `buildResponse` for all error responses so that the correct headers are applied.


zigtan (approved):
Thanks @PurcyTwoBulls !

PR #280, opened 2024-09-02
fix(EG-571): add manifest name underneath run name

simondib:
Small fix to add manifest name underneath run name in the Labs workflow page


zigtan (approved):
Thanks @simondib

PR #286, opened 2024-09-10
feat: cognito password strength check during build

PurcyTwoBulls:
We check the password strength per the cognito spec before deploying to prevent failed deployments.


zigtan (approved):
Thanks @PurcyTwoBulls .

PR #291, opened 2024-09-13
feat(EG-681): add VPC support

zigtan:
This PR adds support for setting up a VPC (Virtual Private Cloud) with Public and Private subnets across two Availability Zones for redundancy.

It also adds the S3 and DynamoDB Gateway Endpoint for the Private subnet, and adds support for VPC peering with another VPC which can be from another AWS Account and Region.

To enable VPC Peering, the `easy-genomics.yaml` configuration file has an added optional BE setting `vpc-peering`:

```
          # Back-End specific settings
          back-end:
              test-user-email: 'demo.user@easygenomics.org'
              test-user-password: '...'
              vpc-peering:
                external-vpc-id: 'vpc-0a8c22...'
                external-aws-account-id: '1234567890'
                external-aws-region: 'ap-southeast-1'
                external-role-arn: 'arn:aws:iam::1234567890:role/AcceptVpcPeeringRole'
                external-cidr-block: '10.12.0.0/16'
```

This is future planning infrastructure work that will allow this project's deployment to integrate with other existing AWS environments if needed.


simondib (approved):
Nice work @zigtan

PR #293, opened 2024-09-17
Feat/eg 659 read workflow reports

PurcyTwoBulls:
Gets the list of results from nf tower


zigtan (approved):
Thanks @PurcyTwoBulls , looks good.

PR #296, opened 2024-09-19
feat(EG-700): make s3 url editable input

simondib:
This PR makes the `input` field editable in the Edit Parameters step, providing flexibility to input an existing s3 bucket URL.

The "Download sample sheet" button has also been moved from Step 3 to Step 2, and will display once the file pairs have been successfully uploaded.


zigtan (approved):
Thanks @simondib

PR #297, opened 2024-09-20
feat(EG-579): add cancel workflow UI functionality

willholt-tb:


zigtan (approved):
Thanks @willholt-tb .

PR #298, opened 2024-09-23
fix(EG-701): pipeline param inputs not updating in store

simondib:
- Adds `watchEffect` on props values to fix the `params` store object not updating. 
- Cache the store as a local const for perf improvement.


zigtan (approved):
Thanks heaps @simondib !

PR #299, opened 2024-09-24
feat(EG-704): update list-organization API for SysAdmin/User filtering

zigtan:
This PR modifies the `/easy-genomics/organization/list-organizations` API to return a list of Organizations depending on the Cognito User Session.

If the User is the SystemAdmin it will return all existing Organizations in the Platform.

If the User is a normal User then it will only return a filtered list of Organizations that the User has been invited to.

### DIFF
--- a/packages/back-end/src/app/controllers/easy-genomics/organization/list-organizations.lambda.ts
+++ b/packages/back-end/src/app/controllers/easy-genomics/organization/list-organizations.lambda.ts
@@ -26,9 +26,6 @@ export const handler: Handler = async (
       return buildResponse(200, JSON.stringify(response), event);
     } else {
       const organizationIds = getOrganizationAccessOrganizationIds(event);
-      if (organizationIds.length === 0) {
-        throw new Error('Unauthorized access');
-      }
 
       const response: Organization[] = await organizationService.list();
       const userOrganizations: Organization[] = response.filter((org: Organization) =>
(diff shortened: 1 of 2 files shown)


packages/back-end/src/app/controllers/easy-genomics/organization/list-organizations.lambda.ts
```diff
@@ -2,16 +2,41 @@ import { Organization } from '@easy-genomics/shared-lib/src/app/types/easy-genom
...
   try {
-    const response: Organization[] = await organizationService.list();
-    return buildResponse(200, JSON.stringify(response), event);
+    if (validateSystemAdminAccess(event)) {
+      const response: Organization[] = await organizationService.list();
+      return buildResponse(200, JSON.stringify(response), event);
+    } else {
+      const organizationIds = getOrganizationAccessOrganizationIds(event);
+      if (organizationIds.length === 0) {
+        throw new Error('Unauthorized access');
```
willholt-tb, line 30: by my reckoning this could be a valid case (user is removed from all organizations but can still log in and do some things) so it seems overzealous to me to return an error - rather simply returning an empty list should suffice so the website shows no organizations but otherwise it's business as usual. thoughts @zigtan ?

zigtan: @willholt-tb , that's true. I'll remove this logic check.

PurcyTwoBulls (approved):

willholt-tb (approved):

PR #300, opened 2024-09-24
feat: unified error response and custom errors

PurcyTwoBulls:
The objective if this change is to:
- allow errors to have proper http response codes (currently hard coded to 400 for all errors)
- allow errors to have an error code that the FE can rely on
- allow BE to execute different logic based on error types instead of attempting string matches
- catch errors from where they are thrown, particularly when they might be thrown from dynamodb
- reduce duplicated code in handlers

About 90% of errors thrown through the handlers have been updated as part of this PR


zigtan (approved):
Thanks @PurcyTwoBulls for this massive enhancement.

Great work!

PR #306, opened 2024-09-25
fix(EG-723): enhance update-laboratory API to support S3 bucket modification

zigtan:
This PR updates the `/easy-genomics/laboratory/update-laboratory` to allow it to modify the S3Bucket attribute for a Laboratory.

This PR also modifies the Laboratory `WorkspaceId` & `AccessToken` verification workflow to perform the verification process without requiring the User to specify the `AccessToken` each time a Laboratory is updated. 

* When the `AccessToken` is not specified as part of the Laboratory update, the `/nf-tower/connection/request-connection-test` will lookup the existing `AccessToken` to verify the `WorkspaceId`.
* Otherwise, when a new `AccessToken` is specified then it will be used to verify the `WorkspaceId`.
* Only when the `WorkspaceId` & `AccessToken` verification workflow passes, then the Laboratory's details will be updated.

Technically, I would prefer the `/nf-tower/connection/request-connection-test` API to be depreciated and have the `WorkspaceId` & `AccessToken` verification workflow logic simplified and integrated into the existing `/easy-genomics/laboratory/update-laboratory` API to make it easier to maintain.

However, there is a bit of FE Lab update UI logic that would need to be updated. This refactoring work can be done in a separate ticket later so as not to block @simondib 's current work to add S3Bucket selection/modification for a Laboratory.


simondib (approved):
Thanks @zigtan!

PR #307, opened 2024-09-25
fix(EG-690): fix pipeline run navigate-out confirmation logic

willholt-tb:


packages/front-end/src/app/pages/labs/[labId]/[pipelineId]/run-pipeline.vue
```diff
@@ -64,42 +60,30 @@
     usePipelineRunStore().setUserPipelineRunName('');
     usePipelineRunStore().setParams({});
     initializePipelineData();
-    resetStepperKey.value++;
```
Comment (simondib) [inline packages/front-end/src/app/pages/labs/[labId]/[pipelineId]/run-pipeline.vue:67]: Was this intentionally removed? It triggers a component reset to clear previous data.

Comment (willholt-tb) [reply packages/front-end/src/app/pages/labs/[labId]/[pipelineId]/run-pipeline.vue:67]: hmmmmm, I removed it because it was mapped to `:key="resetStepperKey"`, but `EGRunPipelineStepper` doesn't have a `key` prop:

```
 const props = defineProps<{
    schema: object;
    params: object;
  }>();
```

So I figured it was dead code that wasn't doing anything. Have I misunderstood?

How would `EGRunPipelineStepper` (mis)behave if it wasn't doing it's job?

Comment (willholt-tb) [reply packages/front-end/src/app/pages/labs/[labId]/[pipelineId]/run-pipeline.vue:67]: I'll have a closer look at how it's behaving in a little bit

Comment (simondib) [reply packages/front-end/src/app/pages/labs/[labId]/[pipelineId]/run-pipeline.vue:67]: `:key` is an attribute, not a prop so it isn't declared inside the component. 

If there's a change to the data property assigned to `key`, ie `resetStepperKey.value++` it will reset the component to its initial state. 

Without it, if a user ops to run another pipeline at the success screen (see line 230 in `EGRunPipelineStepper.vue`) the Stepper will be stuck in its current workflow.

Comment (willholt-tb) [reply packages/front-end/src/app/pages/labs/[labId]/[pipelineId]/run-pipeline.vue:67]: I've put `resetStepperKey` back in like it was before 👍

simondib (approved):

PR #308, opened 2024-09-26
feat(EG-671): select s3 bucket in Lab details page

simondib:
This PR adds the ability to select an S3 bucket during Lab creation or post-creation when editing a Lab.

When creating a lab the list item will default to the first array item from the `/list-buckets` endpoint. 

A new API module named `infra` has been introduced to handle the S3 request and any future infrastructure-related services.


zigtan (approved):
Thanks @simondib !

PR #311, opened 2024-09-30
fix(EG-718): hide params section if its items are all hidden

simondib:
Checks if all param properties in a given section are flagged as `hidden` and if true hides their parent section heading from the sidebar UI.


willholt-tb (approved):
lgtm

PR #312, opened 2024-09-30
feat(EG-573): make submitted pill grey

willholt-tb:
https://dept-au.atlassian.net/browse/EG-573


willholt-tb:
New colours for Submitted pill:
image

PR #315, opened 2024-10-01
fix(EG-502): make grant access buttons load individually instead of all together

willholt-tb:


simondib (approved):
Thanks - this has been a bugbear for a while 😅

PR #317, opened 2024-10-01
feat(EG-496): add option to remove user from lab from edit user page

willholt-tb:


zigtan (approved):
Thanks @willholt-tb !

PR #323, opened 2024-10-02
feat(EG-507): require org admin to create a lab

willholt-tb:


willholt-tb:
this work is more complicated than I thought and will need more attention

PR #325, opened 2024-10-02
feat(EG-738): playwright invite and remove Orguser

marvinsumali:
These are new script for inviting a user to an org and immediately remove it to set it back to original state


marvinsumali:
Im not sure how to remove Simon from the approvers :)

willholt-tb (approved):
lgtm 👍

PR #326, opened 2024-10-03
feat(EG-675): update CDK version & add cdk-nag support

zigtan:
This PR adds the cdk-nag library to help audit the FE and BE CDK infrastructure against security guidelines.

At the moment the implementation only enables the AWS Security check for now, and it only works when the following BE or FE script is run: `pnpm run cdk-audit`.

Our current CDK infrastructure implementation does not fully comply with the AWS Security guidelines, and there will be a lot of errors to resolve. 

So I have only enabled the check when the `pnpm run cdk-audit` is used rather than impose all these errors on the rest of the team and slow down development.

I did make some immediate improvements to the FE CDK implementation based on the auditing and these include:
 - adding a new S3 bucket for access logging
 - enabling CloudFront distribution access logging
 - enabling www S3 bucket server access logging
 - enabling CloudFront distribution GeoRestriction to AU and US regions

Also as part of the introduction of cdk-nag, I updated the project:
 - CDK version to the latest 2.160.0 (after resolving a compatibility issue preventing CDK from updating).
 - PNPM to the latest 9.11.0
 - replaced the depreciated S3Origin object with S3BucketOrigin

**NOTE:** The text input error I was encountering was tracked down to a bug in `@nuxt/ui` library `2.0.18.5` and `2.0.18.6`. I have locked this library to version `2.0.18.4` until the bug can be resolved in a future release.


willholt-tb (approved):
lgtm

PR #330, opened 2024-10-04
feat(EG-734): update nextflow apis to support empty workspace

zigtan:
This PR updates the `/nf-tower/...` APIs to support an empty workspaceId since it is now optional, which means it will default to the Personal Workspace of the owner of the AccessToken.

- it removes previous logic to throw the `LaboratoryWorkspaceIdUnavailableError` exception.
- improves the setting of query parameters for the `/nf-tower/...` APIs since the workspaceId can be empty

This PR also updates the project packages for any missed recent updates.


packages/back-end/src/app/controllers/nf-tower/workflow/read-workflow.lambda.ts
```diff
@@ -78,3 +72,11 @@ export const handler: Handler = async (
     return buildErrorResponse(err, event);
   }
 };
+
+function getApiQueryParameters(event: APIGatewayProxyWithCognitoAuthorizerEvent, workspaceId?: string): string {
```
willholt-tb, line 76: since there's so many uses of this function would it be better as a util?

packages/back-end/src/app/controllers/nf-tower/workflow/read-workflow.lambda.ts
```diff
@@ -78,3 +72,11 @@ export const handler: Handler = async (
...
+  const apiParameters: URLSearchParams = getApiParameters(event);
+  if (workspaceId && workspaceId !== '') {
```
willholt-tb, line 78: a nitpick, but `''` is falsy so `workspaceId !== ''` is redundant

willholt-tb (approved):

PurcyTwoBulls (approved):
Looks good, although that getApiQueryParameters would probably be better in a util than being duplicated in each call.

PurcyTwoBulls (approved):
LGTM

PR #338, opened 2024-10-08
fix(EG-750): setup shared s3 bucket with cors settings for deployment

zigtan:
This PR updates the `data-provisioning-nested-stack.ts` to create a single shared S3 Bucket named `${this.props.env.account!}-${this.props.namePrefix}-lab-bucket` to simulate an existing S3 Bucket for a Lab to select and use.

e.g. `{AWS ACCOUNT ID}-dev-quality-lab-bucket`

It also removes an unnecessary `s3:CreateBucket` IAM policy for the `create-laboratory` API since we are no longer provisioning new S3 Buckets at Lab creation.


PurcyTwoBulls (approved):
LGTM

simondib (approved):

PR #340, opened 2024-10-08
fix(EG-748): update www-hosting-construct to improve cloudfront dist logging

zigtan:
This PR improves upon the CloudFront distribution logging to make it easier to clean up the deployment without running into non-empty access-log S3 Bucket error.

- It reduces the logging window to 1 day (minimum)
- It also only enables the CloudFront distribution logging for non-Dev deployments (e.g. Demo Pre-Prod).


willholt-tb (approved):
lgtm 👍

PR #343, opened 2024-10-09
feat(EG-758): add request-file-download API

zigtan:
This PR adds the POST `/easy-genomics/files/request-file-download` API and it requires the JSON payload:

```
{
    "LaboratoryId": "<< Laboratory ID >>",
    "Path": "<< File Path / S3 Key >>"
}
```

It returns:
```
{
    "DownloadUrl": "<< Pre-Signed S3 URL >>"
}
```

The `DownloadUrl` is set to expire after 60 seconds.

The LaboratoryId is used to:
1. Verify if the User calling the API has access to the Laboratory
2. Retrieve the S3 Bucket for the Laboratory to generate the Pre-signed S3 URL to obtain the file

---

The existing `/easy-genomics/upload/...` APIs should be refactored to sit in the `/easy-genomics/files/...` path for logical grouping. This will need to be done in a separate ticket / PR.

Additionally, the `validateLaboratoryManagerAccess()` and `validateLaboratoryTechnicianAccess()` validation helper functions can be refactored and simplified to something:

```
export function validateLaboratoryUserAccess(event: APIGatewayProxyEvent, organizationId: string, laboratoryId: string, isLabManager?: boolean, isLabTechnician?: boolean) {
  ...
}
```

Again this can be done in a separate ticket / PR.

### DIFF
--- a/packages/back-end/src/app/controllers/easy-genomics/files/request-file-download.lambda.ts
+++ b/packages/back-end/src/app/controllers/easy-genomics/files/request-file-download.lambda.ts
@@ -1,0 +1,75 @@
+import { buildErrorResponse, buildResponse } from '@easy-genomics/shared-lib/lib/app/utils/common';
+import {
+  InvalidRequestError,
+  LaboratoryBucketNotFoundError,
+  UnauthorizedAccessError,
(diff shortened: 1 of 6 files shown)


simondib (approved):
Thanks @zigtan

PR #345, opened 2024-10-10
fix: incorrect operator was used for user access

PurcyTwoBulls:
I made a mistake with the user access when adding lab manager restrictions. The validation check should be failing if neither of the requirements are met, not either of them.


zigtan (approved):
Thanks @PurcyTwoBulls !

PR #346, opened 2024-10-10
feat(EG-454): view and download run results

simondib:
This PR adds the ability to view and download the resulting files from a workflow run. 

**Note:** while the file path (`externalPath`) passed to the download request is correct, there is further back-end need for this feature to be consistently reliable.

`EGPageHeader` also refactored to enforce an explicit back path, as well as accomodate loading states via a configurable  skeleton prop object.


packages/front-end/src/app/components/EGPageHeader.vue
```diff
@@ -1,17 +1,21 @@
...
+    description?: string;
+    showBack: boolean;
+    backButtonLabel?: string;
+    backAction?: () => void;
+  }>();
+
+  const router = useRouter();
+
+  const defaultBackAction = () => {
+    router.go(-1);
```
willholt-tb, line 13: navigating back causes some unintended behaviour if the page is navigated to from an unexpected source. for a quick example try:

- open new tab
- navigate to http://localhost:3000/labs/bbac4190-0446-4db4-a084-cfdbc8102297/4L3Qt66Y9SS4Y7?tab=Run+Details
- click back - it won't do anything or it will go to your new tab homepage, depending on browser

and then try
- navigate to http://localhost:3000/labs/bbac4190-0446-4db4-a084-cfdbc8102297/4L3Qt66Y9SS4Y7?tab=Run+Details
- switch between the tabs on the page a couple times
- click back
- it will switch between tabs because that action changes the url and adds a page to the router stack

I think we need to avoid popping pages from the router to go back, and push the intended back-to page instead - this is what I had to do to fix back-button-related problems last sprint

packages/front-end/src/app/stores/labs.ts
```diff
@@ -36,7 +36,9 @@ const useLabsStore = defineStore('labsStore', {
 
   getters: {},
 
-  persist: true,
+  persist: {
+    storage: localStorage,
```
willholt-tb, line 40: what's the reason for this change? it might affect the labs storage work I'm doing now

simondib: The Nuxt persisted plugin uses cookies as the default storage mechanism, but this was failing to persist the `workflow` object - I suspect it's either a bug with the plugin or an undocumented constraint. Changing it to `localStorage` solves this problem.

This shouldn't impact your work but let me know if you run into any troubles. I'll raise another PR to make this storage change consistent across all stores.

willholt-tb: yeah, good idea to do that, after I posted this comment I ran into the same thing in my dev branch and used the same fix.

packages/front-end/src/app/components/EGPageHeader.vue
simondib: Yep good call, I've made the action empty by default which can be overridden by an explicit path.

packages/front-end/src/app/components/EGPageHeader.vue
```diff
@@ -1,17 +1,36 @@
...
+      backAction: () => {},
       description: '',
-      showBack: true,
     },
   );
+
+  const router = useRouter();
+
+  const skeletonTitleLines = computed(() => props.skeletonConfig?.titleLines ?? 1);
+  const skeletonDescriptionLines = computed(() => props.skeletonConfig?.descriptionLines ?? 1);
```
willholt-tb, line 33: is this a double up on defaults when there's already default values in the props definition?

willholt-tb (approved):
lgtm

PR #350, opened 2024-10-14
refactor(EG-732): rework store caching and data sharing between components

willholt-tb:
I was going to wait to open this PR until the new pipeline refactor was ready, but that's going to be a fair bit of work so I'm opening just this bit of work now.

# Main changes

Workflows are now cached into `pipeline-run` store. They're stored as objects indexed by labid and workflowId for individual lookup on the view workflow page, and for each lab an ordered list of workflow ids is also stored for populating the list of workflows for that lab. Both of these pages load their page contents from the store and are able to trigger updates of the data.

A bit of type refactoring has been done on the api to fix up mistakes regarding the schemas of the results from the api.

`/labs/[labId]/[pipelineId]/run-pipeline` now gets a query param `?workflowTempId={uuid}` to use as a key to access the pinia data for that wip workflow run. This allows pipeline runs to be configured in different tabs independently of each other.

`pipelineRunStore` renamed to `workflowStore`, terminology generally standardized

Labs and Orgs also cached in store

Refactor out most remaining global state singletons

Replace uiStore.isRequestPending bool with set of strings to represent different pending request, allowing concurrent and overlapping loading states to not affect each other

Convert store read functions to getters


willholt-tb:
Given that @simondib is off today I'll combine all of the workflow work into this PR.

packages/front-end/src/app/stores/user.ts
simondib, line 1: I've noticed there's a bunch of actions that should probably be moved to the getters bracket. Could you move `isSuperUser()` and all the functions below it that return booleans? They'll need to be converted to the getter syntax.

This should hopefully also fix the bunch of red squigglys.

packages/front-end/src/app/components/EGRunPipelineFormRunDetails.vue
```diff
@@ -30,7 +31,14 @@
    * viralrecon-illumina_community-showcase_20240712_5686910e783b4b2
    */
   const MAX_TOTAL_LENGTH = 80;
-  const maxRunNameLength = ref(MAX_TOTAL_LENGTH - props.pipelineName.length - 8 - 15 - 2);
+  const maxRunNameLength = computed<number>(() => {
```
simondib, line 34: Just be mindful a change when in yesterday that will override this when you resolve the merge conflicts - it should set the value at a constant 50 characters.

packages/front-end/src/app/stores/workflow.ts
simondib, line 1: As it's a new concept, can you add a comment explaining the idea behind wip workflow for future reference?

packages/front-end/src/app/stores/ui.ts
```diff
@@ -1,31 +1,67 @@
 import { defineStore } from 'pinia';
 
+type PendingRequest =
```
simondib, line 3: Great stuff 👍

packages/front-end/src/app/stores/orgs.ts
```diff
@@ -1,57 +1,47 @@
...
-          firstName: this.selectedUser?.FirstName || '',
-          lastName: this.selectedUser?.LastName || '',
-          email: this.selectedUser?.UserEmail || '',
-        }),
-      );
+    async loadOrgs(): Promise<void> {
+      const { $api } = useNuxtApp();
+      const freshOrgs: Record<string, Organization> = {};
+
+      for (const org of await $api.orgs.list()) {
```
simondib, line 25: `TS18046: $api is of type unknown`

packages/front-end/src/app/composables/useAuth.ts
simondib, line 1: There's a few TS errors in here to address.

simondib (commented):
Nice work. Just a bit of cleanup needed before merge.

packages/front-end/src/app/components/EGRunPipelineFormRunDetails.vue
willholt-tb: just merged and have preserved the 50 char constant 👍

packages/front-end/src/app/stores/workflow.ts
willholt-tb: ✅

packages/front-end/src/app/stores/orgs.ts
willholt-tb: where does that error come up? I'm not seeing it in vscode.

packages/front-end/src/app/composables/useAuth.ts
willholt-tb: I don't have any error reports in this file from vscode - I'm guessing they're coming up in Webstorm? is there another way to produce them that you know of?

simondib (approved):

PR #353, opened 2024-10-16
feat(EG-762): increase run name char limit

simondib:
- Increases pipeline run name to 50 chars max
- Also makes some mandatory field names consistently styled.


zigtan (approved):
Thanks @simondib !

PR #361, opened 2024-10-18
fix(EG-732): add parentheses to invoke canManageOrgs

willholt-tb:


willholt-tb:
This should also fix https://dept-au.atlassian.net/browse/EG-801

simondib (approved):

PR #363, opened 2024-10-18
fix(EG-767): restrict superusers to /admin, block regularusers from /admin

willholt-tb:


packages/front-end/src/app/components/EGHeader.vue
```diff
@@ -57,7 +57,7 @@
           </ULink>
           <ULink
             v-if="useUserStore().canManageOrgs"
-            :to="useUserStore().isSuperuser ? '/admin/orgs' : '/orgs'"
+            :to="'/orgs'"
```
simondib, line 60: String can be simply passed as `to="/orgs"`

simondib (approved):
Just having a look at the middleware we should probably extract the route strings as consts for better re-use. Although this can be done in a separate PR.

packages/front-end/src/app/components/EGHeader.vue
willholt-tb: done 👍

PR #365, opened 2024-10-18
fix(EG-802): fix Lab Run Details page regression

zigtan:
This fixes the Lab Run's Details page regression so the details can load properly.


willholt-tb:
this isn't a complete fix because isRequestPending requires an argument, let me have a look and update you

willholt-tb (approved):
@zigtan  good to go

PR #371, opened 2024-10-21
fix(EG-726): unable to load lab directly in new tab because labs store not persisted

willholt-tb:


packages/front-end/src/app/stores/labs.ts
simondib, line 1: This shouldn't be needed as the persist options are configured globally for all stores in the Nuxt config.

simondib: Ignore the above comment - I was mistaken, the Nuxt config is only setting the localStorage as the default storage mechanism. PR is good 👍

simondib (approved):

packages/front-end/src/app/stores/labs.ts
simondib: @willholt-tb The [plugin docs](https://prazdevs.github.io/pinia-plugin-persistedstate/frameworks/nuxt.html#usage) suggest this syntax: 

```
  {
    persist: true,
  }

```

I've noticed some other stores also need updating their syntax but I can do that separate to this PR.

willholt-tb: right, if the nuxt config sets it to localhost by default then in the store we can just `persist: true`, makes sense

simondib (approved):

packages/front-end/src/app/stores/labs.ts
simondib: Yep. Thanks for updating the other stores 👍

PR #378, opened 2024-10-22
feat(EG-779): implement labs view for superuser

willholt-tb:
# Main changes

- added `/admin/labs`, `/admin/labs/[labId]`
- labs and lab pages are componentized into EGLabsList and EGLabView, used in both regular and admin pages

# Future work

- componentization of labs/lab pages is somewhat half baked
  - where regular user and superuser behaviour diverges, some of is it pulled out to props, some of it is handled conditionally within the component
  - where they interact with the route, most of it is pulled out into a prop, but some of it still interacts directly with the route from within the component (eg. tab query param)
- should be improved later when there's more time
- this may mesh naturally with the future work to support multiple organizations
- see https://dept-au.atlassian.net/browse/EG-815


packages/front-end/src/app/components/EGLabDetailsForm.vue
```diff
@@ -456,7 +456,7 @@
         :size="ButtonSizeEnum.enum.sm"
         type="submit"
         label="Edit"
-        :disabled="!useUserStore().canEditLab(useUserStore().currentOrgId)"
+        :disabled="useUserStore().isSuperuser || !useUserStore().canEditLab(useUserStore().currentOrgId)"
```
simondib, line 459: `!useUserStore().canEditLab(useUserStore().currentOrgId)` might be a bit more readable as: 

`const hasEditPermission = !useUserStore().canEditLab(useUserStore().currentOrgId);`

simondib (approved):
Just one minor suggestion, otherwise looks great.

PR #379, opened 2024-10-22
feat(EG-816): create lab script for an org admin

marvinsumali:
Due to an issue in the pipeline not seeing any test for Org Admin, I've consolidated existing and new tests for an Org Admin including the following:
- 01 - Create a Laboratory Successfully
- 02 - Update a Laboratory Successfully
- 03 - Grant access to a user to a Laboratory via Edit User Access Successfully
- 04 - Remove user from a Laboratory Successfully

**Note:** Please ignore the commented lines, this will be a guide on missing test scripts


simondib (approved):
Nice job @marvinsumali

PR #380, opened 2024-10-22
feat(EG-776): verify current user org access for list endpoints

PurcyTwoBulls:
Users who have been given/removed access for Organizations and Laboratories won't have their org access in their ID Token updated to reflect that unless they log out and log back in.

Here we will check on the Org and Lab list endpoints if that org access is current and throw an error to let the FE know to refresh the users ID Token and try again.


zigtan (approved):
@PurcyTwoBulls , this is great! Nice work.

PR #386, opened 2024-10-23
feat(EG-786): add NextFlow file download support

zigtan:
This PR refactors/enhances the previous pre-signed S3 Download URL API `request-file-download` to `/easy-genomics/file/request-file-download-url` so it can be used for downloading files within the AWS S3 Buckets accessible by the Easy Genomics deployment (e.g. download sample-sheet). 

This API will be used for file downloads outside of NextFlow Tower (i.e. Easy Genomic uploaded files or AWS HealthOmics output files).

---

This PR adds a new NextFlow Tower specific `/nf-tower/file/request-file-download` API to enable Easy Genomics to securely download NextFlow Tower Run Result output files regardless of AWS S3 Bucket or Region.

This is achieved by leveraging NextFlow's generated content URL that consists of the `basePath` and each file's respective `path` contained in the `/nf-tower/workflow/read-workflow-reports/{WorkflowID}` API response.


simondib (approved):
🏆

PR #387, opened 2024-10-23
fix(EG-811): update organization-service delete() to correctly remove unique reference

zigtan:
This PR fixes the organization-service delete() method to correctly remove the Organization's lower-case name from the DynamoDB unique-reference table.


PurcyTwoBulls (approved):

PurcyTwoBulls:
LGTM

PR #389, opened 2024-10-24
fix(EG-808): add store helper to check if can add lab user

simondib:
- Checks in the Lab view if a user is an Org Admin or Lab Manager. 
- Hides "Add Lab Users" button for `superuser` (Admin view)
- Reduce toast timer globally to 3 seconds to improve UX


zigtan (approved):
Thanks @simondib , lgtm.

PR #391, opened 2024-10-25
fix: use email address to find user instead of cognito id

PurcyTwoBulls:
Finds the user from their email address instead of the cognito id, which might not match up to the user inside dynamodb


zigtan (approved):
Thanks @PurcyTwoBulls .

PR #392, opened 2024-10-25
feat(EG-770): add s3 bucket tags to enable bucket filtering for list-buckets API

zigtan:
This PR adds S3 Bucket tagging to:

| Bucket | Tags |
|--------|------|
| **FE www hosting bucket** | `'easy-genomics:s3-bucket-type': 'www'` |
| **BE provisioned lab bucket** | `'easy-genomics:s3-bucket-type': 'data'` |

The `list-buckets` API has been enhanced to lookup the S3 Bucket's Tag details and then filter for only the S3 Buckets containing the Tag key/value pair: `'easy-genomics:s3-bucket-type': 'data'`.

This will allow customers to create their own S3 Buckets / use their existing S3 Buckets with Easy Genomics if they belong to the same AWS Region as the Easy Genomics deployment and have added the tag `'easy-genomics:s3-bucket-type': 'data'`.

AWS S3 prevents looking up an S3 Bucket's tags if the Bucket does not belong to the same AWS Region as the caller (i.e. Easy Genomics deployment). 

As a result, the inability to read an S3 Bucket's tag details can be used as a proxy to identify S3 Buckets in the same AWS Region.

### DIFF
--- a/packages/back-end/src/app/controllers/easy-genomics/list-buckets.lambda.ts
+++ b/packages/back-end/src/app/controllers/easy-genomics/list-buckets.lambda.ts
@@ -1,4 +1,4 @@
-import { Bucket, ListBucketsCommandOutput } from '@aws-sdk/client-s3';
+import { Bucket, GetBucketTaggingCommandOutput, ListBucketsCommandOutput, Tag } from '@aws-sdk/client-s3';
 import { S3Bucket } from '@easy-genomics/shared-lib/src/app/types/easy-genomics/s3-bucket';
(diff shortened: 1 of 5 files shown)


simondib (approved):

PurcyTwoBulls (approved):
LGTM

zigtan:
Added an extra commit to remove the `'easy-genomics:s3-bucket-region'` tag, so customers only need to specify `'easy-genomics:s3-bucket-type': 'www'` for their own S3 Buckets for them to be listed by the `list-buckets` API.

PR #393, opened 2024-10-27
feat(EG-828): make ci/cd test execution failover

simondib:
Makes the `test-e2e` command failover so that if any of the test scripts fail the next script will try and execute. 

e.g. Currently tests for Sys Admin run before Org Admin. If any of the Sys Admin tests fail, the entire script exists, preventing Org Admin tests from running.


marvinsumali (approved):
Approved!

PR #394, opened 2024-10-28
feat(EG-821): create labs under test organization

marvinsumali:
This PR includes updates to existing Labs scripts as Org Admin
- It deletes the user first before deleting a Lab
- Deletes Labs first before creating a new one
- Error handling to check first if records exists before performing actions
- Includes new scripts for: Granting access via Edit User Access, adding User to a Lab

**Note:** We cannot create Labs yet out of a newly created ORG due to not having automated scripts for accepting invites so for the meantime we are using the **'Default Organization'** when creating the Labs


packages/front-end/tests/e2e/org-admin/Labs_Org_Admin.spec.e2e.ts
```diff
@@ -1,34 +1,170 @@
 import { test, expect } from 'playwright/test';
+//const invitemail = 'marvin.umali+eg99@deptagency.com';
+//const orgName = 'Automated Org';
```
simondib, line 3: Is there a different email that can be added here rather than your personal one? 

Also may be better to add these as // TODOs with an accompanying comment about the intent

marvinsumali: Unfortunately we don't have a group email yet. Maybe it has to be requested so that few members will receive an email (e.g. team@easygenomics.org).
OK will add to do for the comments

zigtan (approved):
Thanks @simondib , @marvinsumali , LGTM.

PR #395, opened 2024-10-28
feat(EG-781): restrict access to nf-tower apis to org admins and lab …

PurcyTwoBulls:
Restricts access to all of our NF Tower related api's


zigtan (approved):
Thanks @PurcyTwoBulls , LGTM.

PR #396, opened 2024-10-29
fix(EG-821): e2e test improvements

simondib:
Various improvements and fixes for `Organizations_System_Admin.spec.e2e.ts`:

- adds `exact: true` option for string specificity
- camelCase variables for consistency
- add uuid to randomize string to prevent test breakage due to previous test failures


marvinsumali (approved):
lgtm, approved!

marvinsumali:
Have updated the email used for testing. I think this is good to go

PR #398, opened 2024-10-29
fix(EG-836): remove SES IAM restrictions on email identities to enable broader end-user testing

zigtan:
This PR removes the SES IAM restrictions for email identities `'*@twobulls.com'`, and `'*@deptagency.com'` to allow for broader testing, and also enable customers to invite users to their own deployment properly.


simondib (approved):
Thanks @zigtan!

PR #399, opened 2024-10-30
feat(EG-712): add accordion ui to stepper (step 4)

simondib:
Adds accordions (default to open) with expand/contract toggle button to Step 4 of the Run Pipline stepper flow to improve UX.


zigtan (approved):
Thanks @simondib , works great!

PR #400, opened 2024-10-30
fix(EG-821): specific selector + better logging

simondib:
Adds more specificity to the Admin table selectors plus some additional logging for easier future debugging.


marvinsumali (approved):
tested and looks good!

PR #403, opened 2024-11-06
feat(EG-777): refresh expired token on error EG-110

simondib:
This PR aims to seamlessly refresh the UI/user state to reflect a change to the user's permissions while the user is signed into the system. 

An example of this is if an Org Admin is signed in, but in another browser session a Sys Admin revokes the Org Admin's access. In this case the next API request to `/organization/list-organizations` will fail and update the UI to reflect this accordingly (e.g. remove the Org menu item and prevent). 

The frontend achieves this by listening for an `EG-101` error code in any API request, and if found requests a refreshed auth token, refreshes the user permissions then remounts the Nuxt app (via an incremented remount key on `NuxtLayout`) so as to refresh the UI to match the updated user permissions.


zigtan (approved):
Thanks @simondib , this is awesome!

PR #405, opened 2024-11-06
fix(EG-818): org details disabled/loading states while editing

simondib:
Disables input fields and adds a loading/disabled state to the "Save changes" button when editing Org details.


zigtan (approved):
Thanks @simondib .

PR #406, opened 2024-11-07
feat(EG-841): download actual samplesheet

zigtan:
This PR updates the `Run Pipeline Step 2` UI's `Download sample sheet` button logic to leverage the `/easy-genomics/file/request-file-download-url` API to retrieve the current samplesheet.csv file store on S3.

This replaces the previous implementation which obtained a copy of the CSV content from the BE API `/easy-genomics/upload/create-file-upload-sample-sheet` API which is only called once after all the files are uploaded.

The use of the `/easy-genomics/file/request-file-download-url` API allows the UI to reliably download the the latest samplesheet.csv file from S3. This will complement the planned `Run Pipeline Step 2` UI's enhancement to allow the end-user to also upload their own samplesheet.csv file to S3.


simondib (approved):
Nice job @zigtan

PR #410, opened 2024-11-11
feat(EG-825): Create 'Labs' tab within Organizations page for System Admin

simondib:
Replaces the existing Labs menu item with a Labs tab under the Orgs menu item for a System Admin / Admin view.

This PR also includes related fix + feature work:

- Fix: EG-822 - prevents Sys Admin from removing a lab
- Feat: EG-800 - adds Admin View chip


zigtan (approved):
Thanks @simondib !

PR #412, opened 2024-11-11
fix(EG-533): improve create-user-invitation-request & confirm-user-invitation-request API error handling

zigtan:
This PR fixes two bugs with the user invitation workflow:

1. When a User is invited more than once, the Cognito generated encrypted temporary password embedded in the Invitation's JWT is invalided. So it leads to the temporary password test login to fail with an error `"Incorrect username or password"` message.

The fix for this is to catch the error from the `Cognito-IDP service's initiateAuth()` and throw an appropriate error message for the FE to handle. (This is not currently done yet, and will need the FE team to complete). But the error message is logged in the Browser's console.log for now, and in the AWS Cloudwatch logs.

2. The successful activation of a User, and subsequent removal from an Organization cannot receive re-invitation emails from the Cognito trigger. There were two solutions to this problem:
  a) Send an existing User email invitation using SES directly and require the User to complete the same invitation UI form again (despite not changing their User name details or Password)
  b) Send an existing User email invitation courtesy notification email using SES directly and just automatically set their OrganizationUser Status to 'Active'.

Option `b` was chosen as it was the most user friendly solution.


simondib (approved):
👍

packages/shared-lib/src/app/utils/HttpError.ts
```diff
@@ -31,17 +31,17 @@ export class NotFoundError extends HttpError {
...
 }
 
 /**
  * Invalid body parameters request
  */
 export class InvalidRequestError extends HttpError {
-  constructor() {
-    super('Invalid request', 400, 'EG-102');
+  constructor(message?: string) {
+    super(message ? `Invalid request: ${message}` : 'Invalid request', 400, 'EG-102');
```
PurcyTwoBulls, line 44: Maybe the super could take the optional param of perhaps `details` and if present it is appended to the end of the message, that would save a lot of duplicate-ish code in each error type

PurcyTwoBulls (commented):
Just some queries around the error messaging, otherwise looks good.

PR #414, opened 2024-11-12
fix(EG-845): lab manager technician access validation

PurcyTwoBulls:
The lab manager and technician were incorrect, but usually cancelled each other out when used together to give the appearance they were working, but for instances were only one is used they will fail. Have fixed them and added tests to prevent issues in the future.

Also the add lab user request was expecting an organization id that is not currently supplied by the FE, and that was not required by the BE code.


packages/shared-lib/src/app/schema/easy-genomics/laboratory-user.ts
```diff
@@ -19,7 +19,6 @@ export const AddLaboratoryUserSchema = z
   .object({
     LaboratoryId: z.string().uuid(),
     UserId: z.string().uuid(),
-    OrganizationId: z.string().uuid(),
```
zigtan, line 22: This is required, and will cause a regression. We store the OrganizationId when adding a LaboratoryUser mapping record:

https://github.com/twobulls/easy-genomics/commit/e0c5e045020b82d74e21e2e00c7d36c296f5c6f7#diff-685fc3a899953bce9053044c7fe08639ca372f77f06b57b1617c17e2a4b9b476R1000

PurcyTwoBulls: Ah ok, I see what you mean. Currently (or at least when I did my last FE build the other day) the FE doesn't pass the org so the api call was failing.

I think I can fix this without needing the Org Id from the FE.

zigtan: I can see where you are going. I've added a commit to update the FE so that it doesn't need to pass the OrganizationId across so that we can leverage the Laboratory's existing OrganizationId reference to add to the LaboratoryUser mapping record.

zigtan (approved):

PR #417, opened 2024-11-12
feat(EG-792): create sys admin script for labs

marvinsumali:
This new script for System Admin aims to
- check if the All labs tab is empty and display a console message
- check if Org has a lab and if so click on the lab
- check if the Lab has a user
- check if if the user can access the Edit button in the lab details tab


simondib (approved):
Confirming new test pass locally. Nice work @marvinsumali

PR #419, opened 2024-11-14
fix(EG-840): fix create-user-forgot-password-request API error handling

zigtan:
This PR fixes the `/easy-genomics/user/create-user-forgot-password-request` API so that it always returns a false positive (200 OK response) if it encounters an error from:
* Invalid Forgot Password JWT
* User Account deactivated (Status = `'Inactive'`)
* User Cognito Account deactivated

These errors will continue to be logged in the respective AWS CloudWatch logs for debugging.

But it is important for this API to return a false positive to the FE in order to prevent snooping of User account emails.


PurcyTwoBulls (commented):
LGTM

PurcyTwoBulls (approved):

PR #420, opened 2024-11-15
fix(EG-841): fix request-file-download-url API to handle us-east-1 S3 requests

zigtan:
This PR fixes the `/easy-genomics/file/request-file-download-url` API to generate pre-signed download URLs for the specified download file.

The S3 API `getBucketLocation()` call returns null for S3 Bucket's that are located in `us-east-1`. S3 Bucket's located in other S3 regions will return always return their region value.

The fix improves the  locationConstraint checking for `us-east-1`. If the S3 API response was to be modified in the future to return `us-east-1` instead of null, the logic will still work.


PurcyTwoBulls (approved):
LGTM

PR #421, opened 2024-11-18
fix(EG-863): prevent lab manager from editing lab details

willholt-tb:
Per title, prevent lab manager from editing lab details. Also included some small refactors.

### DIFF
--- a/packages/front-end/src/app/pages/labs/index.vue
+++ b/packages/front-end/src/app/pages/labs/index.vue
@@ -1,12 +1,11 @@
 <script setup lang="ts">
-  const $route = useRoute();
   const $router = useRouter();
 </script>
 
 <template>
   <EGPageHeader title="Labs" :show-back="false">
     <EGButton
-      v-if="useUserStore().canCreateLab()"
+      v-if="useUserStore().canCreateLab(useUserStore().currentOrgId)"
       label="Create a new Lab"
       class="self-end"
       @click="() => $router.push({ path: `/labs/create` })"
--- a/packages/front-end/src/app/pages/labs/index.vue
+++ b/packages/front-end/src/app/pages/labs/index.vue
@@ -5,7 +5,7 @@
 <template>
   <EGPageHeader title="Labs" :show-back="false">
     <EGButton
-      v-if="useUserStore().canCreateLab(useUserStore().currentOrgId)"
+      v-if="useUserStore().canCreateLab()"
       label="Create a new Lab"
       class="self-end"
       @click="() => $router.push({ path: `/labs/create` })"
--- a/packages/front-end/src/app/components/EGHeader.vue
+++ b/packages/front-end/src/app/components/EGHeader.vue
@@ -39,7 +39,7 @@
             Labs
           </ULink>
           <ULink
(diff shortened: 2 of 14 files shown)


packages/front-end/src/app/pages/labs/index.vue
```diff
@@ -1,12 +1,11 @@
 <script setup lang="ts">
-  const $route = useRoute();
   const $router = useRouter();
 </script>
 
 <template>
   <EGPageHeader title="Labs" :show-back="false">
     <EGButton
-      v-if="useUserStore().canCreateLab()"
+      v-if="useUserStore().canCreateLab(useUserStore().currentOrgId)"
```
simondib, line 8: Given the org id isn't ever changing, rather than pass `useUserStore().currentOrgId` as an arg in all of these calls, could it instead be referenced directly in the method? e.g. 

```v-if="useUserStore().canCreateLab()"```

```
    canCreateLab:
      (_state: UserStoreState): boolean =>
        !useUserStore().isSuperuser && useUserStore().isOrgAdmin(useUserStore().currentOrgId)
```

Similarly for canManageOrgs, canEditLabUsers, etc.

willholt-tb: That's what we're coming from. I changed it to anticipate implementing proper org multitenancy, as we will need to specify the org when that happens, so the current status of the org id never changing is a temporary situation. Also, all the other functions below in this file take orgId as a parameter (as relevant) so this change brought these two functions in line with them.

I don't feel too strongly about this, and if we want to use currentOrgId within these permissions functions and not take orgId as a parameter, I'd be inclined to make the complete opposite change and get rid of the orgId parameter in all the other functions where it currently appears, so things are consistent.

Thoughts?

simondib: Ah yep, I remember now. 

How about making the arg optional but default it's value to the current org if not passed in?

willholt-tb: I'm not aware of a good way to do that with Typescript, because optional orgId would come before mandatory labId, which creates a clunky function signature.

I think omitting orgId everywhere is fine too. It makes the code easier for now, and re-adding it will just be a migration task for when we implement multitenancy (there will be migration work for that anyway).

Sound good to you @simondib ?

simondib: > I'm not aware of a good way to do that with Typescript, because optional orgId would come before mandatory labId, which creates a clunky function signature.

Are you referring to `canEditLabUsers`? If so you could just make the optional org id the last arg with a default value.

willholt-tb: Are we ever going to want to specify orgId before multitenancy is implemented anyway? Not sure it adds anything that we will use.

simondib: I think the optional approach gives flexibility, but I'll leave it with you to decide.. it's not a hill I'm willing to die on :sweat_smile:

willholt-tb: Yeah nah, me either, especially as it's a temporary implementation 👍 

We can change it any time if needed.

simondib (approved):

PR #425, opened 2024-11-22
refactor: update instances of "workflow" terminology with "run" where being used to describe a pipeline run

willholt-tb:
In light of the decision that "workflow" = "pipeline", I've updated the many places in the code where "workflow" was used to describe a pipeline run, to prevent future confusion and reduce cognitive load when working on the frontend codebase.

The terminology has been replaced with "run" (or "nextFlowRun" where the implementation is likely to be specific to NextFlow - this involved some assumptions and may be changed when we come to implement Omics functions).

This PR has a lot of lines changed but it should be a strictly uder-the-hood patch with no impact on app behaviour. The only types of changes included are:
- renaming functions, variables, comments, store, etc (should have no impact on code function)
- a couple minor instances of removing dead code that I noticed

I have covered almost all of the frontend code base. The exception is I stopped short of touching the api handler, as the API uses workflow terminology still. This means that the api is still invoked with workflow terminology, eg.

```
const runs: NextFlowRun[] = await $api.workflows.list(labId);
```

This is the crossover point for the terminology.


zigtan:
Thanks @willholt-tb for working on this. 

This somewhat complicated by AWS HealthOmics's own terminology using `Workflows` (equivalent to Seqera / NextFlow Pipelines) and `Runs` (same meaning with Seqera / NextFlow on their web UI - but on Seqera / NextFlow BE APIs they use Workflows).

We will need to discuss with the rest of the team on the best terminology going forward.

zigtan (approved):
Thanks @willholt-tb .

zigtan (approved):

simondib (approved):

PR #429, opened 2024-11-25
feat(EG-830): Run Results file explorer

simondib:
This update replaces the existing flat folder view for Run Results with a dynamic, navigable file/folder structure with breadcrumbs that reflects the hierarchy of the `/easy-genomics/file/request-list-bucket-objects` `Contents` response object.

The response is transformed into a nested structure that's fed into the data table to construct a simulated file system. All files are downloadable via the Actions menu icon. Folder contents will also be downloadable as a zip in a subsequent PR once the backend API has been built. 

**NOTE:** Currently the top level folder is mapped to all the Lab's pipeline runs, rather than the specific run. This is due to the Run ID not currently being available for the FE to append to the `S3Prefix` value as part of the `/easy-genomics/file/request-list-bucket-objects` request.

Additional notes: 

- helpers added from the @vueuse library: `useChangeCase` and `useDebounceFunc`
- add checkbox selection prop option to `EGTable` for selection of row items (this feature was de-scoped from the current ticket but left code in for future use)


zigtan (approved):
Thank you @simondib ! Great work!

willholt-tb (approved):

PR #439, opened 2024-12-02
feat(EG-872): add '/aws-healthomics/run/list-runs' API endpoint

zigtan:
This PR adds the `/aws-healthomics/run/list-runs` API endpoint.

It requires the query parameter `?laboratoryId={LaboratoryId}` to verify if the Lab has AWS HealthOmics access enabled.

It also supports the following optional query parameters:
 - `maxResults`: pagination limit
 - `nextToken`: pagination offset index
 - `name`: search filtering by run name
 - `status`: search filtering by run status 

Note: the  `/aws-healthomics/workflow/list-private-workflows` API does not support the status query parameter so it is explicitly removed to avoid SDK request type errors.


PurcyTwoBulls (approved):
lgtm

willholt-tb (approved):
lgtm

PR #440, opened 2024-12-02
fix: add run roles to lab delete api

PurcyTwoBulls:
Adds the correct role to allow lab deletion to queue up the lab run deletes


zigtan (approved):
Thanks @PurcyTwoBulls , looks good.

PR #445, opened 2024-12-03
feat: user delete api

PurcyTwoBulls:
Allows the deletion of a users account, including their cognito account, laboratory user and organization user.


zigtan (approved):
Thanks @PurcyTwoBulls. Looks good.

PR #452, opened 2024-12-04
feat(EG-874): omics get workflow api

PurcyTwoBulls:
## Title
AWS Omics get workflow api

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
This new endpoint takes in a workflowId (as `id`) in the path and a laboratoryId as a query param. The laboratory is used for checking if the laboratory has access to AWS Omics, and also for user permissions. This is only for private workflows.

## Testing
You will need to have a private workflow already setup in AWS Omics. You can use the _list-private-workflows_ endpoint to list out existing private workflows, then use the id's with this endpoint. If there are no private workflows you will need to create one in AWS, either through the console or the command line.

## Impact
No other changes or impacts to the project outside of added functionality

## Additional Information
Laboratories at present can't be enabled for AWS Omics via the FE, this may change in the future but I had to manually enable this via dynamodb, as well as adding a private workflow.

Not all AWS regions have AWS Omics. If your environment is set up in a region without Omics, you won't be able to use these features.

## Checklist

### DIFF
--- a/packages/back-end/src/app/controllers/aws-healthomics/workflow/read-private-workflow.lambda.ts
+++ b/packages/back-end/src/app/controllers/aws-healthomics/workflow/read-private-workflow.lambda.ts
@@ -1,0 +1,75 @@
+import { GetWorkflowCommandInput } from '@aws-sdk/client-omics/dist-types/commands/GetWorkflowCommand';
+import { buildErrorResponse, buildResponse } from '@easy-genomics/shared-lib/lib/app/utils/common';
+import {
+  LaboratoryNotFoundError,
+  RequiredIdNotFoundError,
+  UnauthorizedAccessError,
+} from '@easy-genomics/shared-lib/lib/app/utils/HttpError';
+import { Laboratory } from '@easy-genomics/shared-lib/src/app/types/easy-genomics/laboratory';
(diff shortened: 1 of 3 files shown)


zigtan (approved):

packages/back-end/src/app/controllers/aws-healthomics/workflow/read-private-workflow.lambda.ts
```diff
@@ -0,0 +1,86 @@
...
+  validateLaboratoryManagerAccess,
+  validateLaboratoryTechnicianAccess,
+  validateOrganizationAdminAccess,
+} from '@BE/utils/auth-utils';
+
+const laboratoryService = new LaboratoryService();
+const omicsService = new OmicsService();
+
+/**
+ * This GET /aws-healthomics/workflow/read-private-workflow/{:id}?workflowId={workflowId}
```
zigtan, line 24: Just noticed this typo comment which should read: `/aws-healthomics/workflow/read-private-workflow/{:id}?laboratoryId={LaboratoryId}`

PR #480, opened 2024-12-18
feat(EG-888): Set a Seqera run S3 folder as default in file explorer

simondib:
## Title*
Points a Seqera run to its associated S3 build folder in the file explorer.

## Type of Change*
- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [X] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [X] UI/UX improvement

## Description
This PR is an enhancement to the existing Seqera Pipeline Run workflow, such that once a workflow is submitted successfully the S3 folder of the run is set as the default folder to view under the Run Results tab.  Prior to this, all S3 folders associated with the Laboratory were listed.

Once data files are uploaded the FE calls `/create-labaratory-run` and includes a newly created uuid as `RunId`. Once the run is successfully launched and a `transactionId` value is returned from Seqera, the FE updates the Laboratory table with this value as `ExternalRunId`. This value corresponds to the build's final S3 path, in the format `s3://{S3 Bucket}/{organizationId}/{LaboratoryId}/next-flow/{transactionId}/`, which is later referenced in the "Run Results" tab when a user views the files/folders of the run.

## Testing*
Manual browser testing + e2e tests passing on dev environment.

## Impact
Optimized time for S3 retrieval, negating most of the file explorer's loading state and time to render to browser.

## Additional Information
None.

## Checklist*


packages/shared-lib/src/app/schema/easy-genomics/laboratory-run.ts
```diff
@@ -18,7 +18,7 @@ export const LaboratoryRunSchema = z
     ExternalRunId: z.string().optional(),
     S3Input: z.string().optional(),
     S3Output: z.string().optional(),
-    Settings: z.string().optional(), // JSON string
+    Settings: z.union([z.string(), z.record(z.string(), z.any())]).optional(), // JSON string
```
simondib, line 21: @zigtan This was necessary in order to validate against both `{}` and `""`values currently being returned in the laboratory runs response.

packages/front-end/src/app/components/EGLabView.vue
```diff
@@ -92,6 +95,8 @@
     };
   }
 
+  const isAllDataLoaded = computed(() => !useUiStore().anyRequestPending(['loadLabData', 'getSeqeraPipelines']));
```
willholt-tb, line 98: There are many loading requests on this page and different components relate to different sets of requests to be considered 'loading' or not.

The changes where `isAllDataLoaded` has replaced existing expressions are not quite right because of this. For example, if the Runs data has loaded, but the Pipelines data is still loading, the Pipelines tab should show loading but the Runs tab should not. This is how it was set up before.

I propose removing `isAllDataLoaded` and reverting the associated changes.

packages/front-end/src/app/components/EGLabView.vue
```diff
@@ -61,6 +63,7 @@
   type GenericRun = {
     type: 'seqera' | 'omics';
     id: string;
+    runId: string;
```
willholt-tb, line 66: How is `runId` different from `id`? Can this be clarified?

packages/front-end/src/app/components/EGRunPipelineFormReviewPipeline.vue
```diff
@@ -53,9 +53,24 @@
           paramsText: paramsText,
         },
       };
-      await $api.seqeraRuns.createPipelineRun(labId, launchRequest);
-      delete runStore.wipSeqeraRuns[seqeraRunTempId];
-      emit('has-launched');
+
+      const res = await $api.seqeraRuns.createPipelineRun(labId, launchRequest);
+
+      if (res) {
```
willholt-tb, line 59: I think it would be better to invert this to the usual pattern and exit early on error:

```
if (!res) {
  throw new Error(...)
}
```

packages/front-end/src/app/pages/labs/[labId]/seqera-run/[workflowName].vue
Comment (willholt-tb) [inline packages/front-end/src/app/pages/labs/[labId]/seqera-run/[workflowName].vue:1]: This url change is confusing me - 

1. the value is definitely an id, not a name 
image
2. to avoid terminology confusion we have `Seqera Pipeline`, `Omics Workflow`, and then `Run`s which can be of the `Seqera Run` or `Omics Run` varieties - as in the previous url path `/seqera-run`
3. this keeps symmetry between seqera and omics runs
image


I don't understand why this would be changed.

packages/shared-lib/src/app/types/easy-genomics/laboratory-run.d.ts
```diff
@@ -30,7 +30,7 @@ import { BaseAttributes, RunType } from "../base-entity";
 export interface LaboratoryRun extends BaseAttributes {
   LaboratoryId: string; // DynamoDB Partition Key (String)
   RunId: string; // DynamoDB Sort Key (String) & Global Secondary Index (String)
-  UserId: string; // Global Secondary Index (String)
+  UserId?: string; // Global Secondary Index (String)
```
zigtan, line 33: The UserId property is intended to efficiently query the LaboratoryRuns by a User, so this should be a required field.

willholt-tb (changes requested):

packages/front-end/src/app/components/EGLabView.vue
simondib: Ah yep, sorry about that, I misread the request names and thought they were all repeated.

packages/shared-lib/src/app/types/easy-genomics/laboratory-run.d.ts
simondib: From memory this was causing a Zod validation error due to the `UserId` not being present, but I can't recall in which request. I can't currently replicate so I'll revert this change.

packages/front-end/src/app/components/EGLabView.vue
simondib: `runId` is redundant so I'll remove.

packages/front-end/src/app/pages/labs/[labId]/seqera-run/[workflowName].vue
Comment (simondib) [reply packages/front-end/src/app/pages/labs/[labId]/seqera-run/[workflowName].vue:1]: Apologies, this was an unintended commit.

packages/front-end/src/app/components/EGLabView.vue
```diff
@@ -635,7 +668,7 @@
       <!-- HealthOmics Pipelines tab -->
       <div v-if="item.key === 'omicsWorkflows'" class="space-y-3">
         <EGTable
-          :row-click-action="viewRunOmicsWorkflow"
+          :row-click-action="(row) => viewRunDetails(row)"
```
willholt-tb, line 671: this isn't right, `viewRunDetails` is for viewing a Run, but this action is on click of an `Omics Workflow` - it should stay as `viewRunOmicsWorkflow` which currently just toasts that it's not implemented yet.

packages/front-end/src/app/components/EGLabView.vue
```diff
@@ -182,11 +184,25 @@
...
   }
 
   function viewRunResults(run: GenericRun) {
-    $router.push({ path: `/labs/${props.labId}/${run.type}-run/${run.id}`, query: { tab: 'Run Results' } });
+    const foundRun = getLabRunId(run.id);
+
+    if (foundRun) {
+      $router.push({
+        path: `/labs/${props.labId}/${run.type}-run/${run.id}`,
+        query: { tab: 'Run Results', runId: foundRun.RunId },
```
willholt-tb, line 203: requiring the (laboratory) `runId` to be passed as a query param will mean that the Run's page is coupled to this click event on the previous page. undoing this pattern has been the cause of a lot of our refactoring time.

the pattern I think we need to follow is to load Laboratory Runs into a store and look them up on the Run page, same as what we do for orgs, labs, pipelines/workflows, and runs.

@simondib I'm aware that it's about 4pm on your last day for the year - I'm happy to get this over the line tomorrow if that's good with you

packages/front-end/src/app/pages/labs/[labId]/seqera-run/[seqeraRunId].vue
```diff
@@ -11,89 +11,78 @@
 
   const labId = $route.params.labId as string;
   const seqeraRunId = $route.params.seqeraRunId as string;
+
   const seqeraRunReports = ref([]);
   const s3Contents = ref<S3Response>(null);
   const tabIndex = ref(0);
+  const runId = ref(''); // Run ID from query params
```
Comment (willholt-tb) [inline packages/front-end/src/app/pages/labs/[labId]/seqera-run/[seqeraRunId].vue:18]: can this be replaced with a `computed` and get rid of the state handling and the `watch` below?

Comment (simondib) [reply packages/front-end/src/app/pages/labs/[labId]/seqera-run/[seqeraRunId].vue:18]: Good call!

packages/front-end/src/app/components/EGLabView.vue
simondib: Hmm, is this a major issue? The page isn't navigable outside of the row link and it avoids having to deal with store state. 

(Having said that I might be overlooking the negative impact of this pattern.. my mental energy is getting very low at this point of the day/week/year 😅 So I'd be very grateful if you could take the refactor on).

packages/front-end/src/app/components/EGLabView.vue
```diff
@@ -182,11 +184,25 @@
   ];
 
   function viewRunDetails(run: GenericRun) {
-    $router.push({ path: `/labs/${props.labId}/${run.type}-run/${run.id}`, query: { tab: 'Run Details' } });
+    const foundRun = getLabRunId(run.id);
+
+    if (foundRun) {
```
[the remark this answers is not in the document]
simondib: This won't be a valid use case for future runs to warrant a toast, but I'll add error handling now.

packages/front-end/src/app/components/EGLabView.vue
willholt-tb: yep all good 👍 

more than a couple of my big refactor tasks have been a flow being built around one entry point and then needing reworking to enable another. my other PR that's open right now is one of these, where the setup for a run is done on click of the pipeline row - then the Run Another button reloads the page but that setup on row click wasn't there, so it had to be pulled into the run page. 

in the longer run these Laboratory Runs should replace the SeqeraRun / OmicsRuns that we already store cache for this reason, so it should end up with a store of its own anyway. might as well do it tomorrow :)

simondib: Perfect, sounds like a worthwhile change. Thanks for the explanation and for taking it on 👍

zigtan (approved):
Thanks @simondib . Approved. We will continue to enhance this implementation whilst you are away.

willholt-tb (approved):

PR #483, opened 2024-12-20
feat: update laboratory run status for nf tower workflow runs

PurcyTwoBulls:
## Title*
Update nf tower workflow run statuses via a delayed queue

## Type of Change*
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
After a NF Tower run is started, check the status of that run and update the laboratory run entry for that run every 5 minutes until the status is `Failed`, `Succeeded` or `Cancelled`. 

## Testing*
After creating a workflow run on NF Tower, there will be a laboratory run entry created for that run (the `ExternalRunId` value will match the workflowId return by NF Tower when starting the run). The status of that laboratory run will be `Pending` initially, but will be updated after 5 minutes to the status of the run in NF Tower. This may currently only be visible via the api in postman.

## Impact
We are now checking these statuses every 5 minutes for incomplete runs. For the expected loads in a production environment we do not expect to push any limits around Lambda invocations, but this will add to an overall load.

## Checklist*

### DIFF
--- a/packages/back-end/src/infra/stacks/nf-tower-nested-stack.ts
+++ b/packages/back-end/src/infra/stacks/nf-tower-nested-stack.ts
@@ -61,44 +61,6 @@ export class NFTowerNestedStack extends NestedStack {
         effect: Effect.ALLOW,
       }),
     ]);
-    this.iam.addPolicyStatements('laboratory-run-id-query-policy', [
-      new PolicyStatement({
-        resources: [
-          `arn:aws:dynamodb:${this.props.env.region!}:${this.props.env.account!}:table/${this.props.namePrefix}-laboratory-run-table`,
-          `arn:aws:dynamodb:${this.props.env.region!}:${this.props.env.account!}:table/${this.props.namePrefix}-laboratory-run-table/index/*`,
-        ],
-        actions: ['dynamodb:Query'],
-        effect: Effect.ALLOW,
-      }),
-    ]);
-    this.iam.addPolicyStatements('laboratory-run-create-policy', [
-      new PolicyStatement({
-        resources: [
-          `arn:aws:dynamodb:${this.props.env.region!}:${this.props.env.account!}:table/${this.props.namePrefix}-laboratory-run-table`,
-          `arn:aws:dynamodb:${this.props.env.region!}:${this.props.env.account!}:table/${this.props.namePrefix}-laboratory-run-table/index/*`,
-        ],
-        actions: ['dynamodb:PutItem'],
-        effect: Effect.ALLOW,
-      }),
-      new PolicyStatement({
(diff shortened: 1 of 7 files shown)


zigtan (approved):
Thanks @PurcyTwoBulls , looks good. Just need to update this branch with the latest main branch.

packages/back-end/src/infra/stacks/nf-tower-nested-stack.ts
```diff
@@ -59,6 +61,16 @@ export class NFTowerNestedStack extends NestedStack {
...
+    this.iam.addPolicyStatements('laboratory-run-id-query-policy', [
+      new PolicyStatement({
+        resources: [
+          `arn:aws:dynamodb:${this.props.env.region!}:${this.props.env.account!}:table/${this.props.namePrefix}-laboratory-run-table`,
+          `arn:aws:dynamodb:${this.props.env.region!}:${this.props.env.account!}:table/${this.props.namePrefix}-laboratory-run-table/index/*`,
+        ],
+        actions: ['dynamodb:Query'],
+        effect: Effect.ALLOW,
+      }),
+    ]);
```
zigtan, line 73: This `'laboratory-run-id-query-policy'` doesn't look to be used anywhere in this stack, so it should be removed.

packages/back-end/src/infra/stacks/nf-tower-nested-stack.ts
```diff
@@ -77,6 +89,16 @@ export class NFTowerNestedStack extends NestedStack {
...
+    this.iam.addPolicyStatements('laboratory-run-update-policy', [
+      new PolicyStatement({
+        resources: [
+          `arn:aws:dynamodb:${this.props.env.region!}:${this.props.env.account!}:table/${this.props.namePrefix}-laboratory-run-table`,
+          `arn:aws:dynamodb:${this.props.env.region!}:${this.props.env.account!}:table/${this.props.namePrefix}-laboratory-run-table/index/*`,
+        ],
+        actions: ['dynamodb:UpdateItem'],
+        effect: Effect.ALLOW,
+      }),
+    ]);
```
zigtan, line 101: Same with this `'laboratory-run-update-policy'` - it doesn't look to be used anywhere in this stack, so it should be removed.

zigtan (approved):
Nice work @PurcyTwoBulls . Just a couple of IAM policies which doesn't seem needed and should be cleaned up.

We may need to review the various type of Status values with the FE team to ensure the displayed Status info is useful for the end-user across both Seqera & AWS HealthOmics.

PR #513, opened 2025-01-21
fix(EG-940): remove lab data check when loading users labs

PurcyTwoBulls:
## Title*
Removed lab data check when loading user labs

## Type of Change*
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
The code that fetches users labs also checks in the labs have loaded, this creates a race condition when the user labs api returns before the org labs are fetched and triggers the "no labs" error to display instead of content.

Have removed the extra check after the user labs call, which doesn't need it.

## Testing*
Refreshing the page multiple times and making sure it doesn't show a "no labs" error. Can also use the network tab in the browser to see if the page still loads when the "list-laboratory-users" endpoint finishes before the "list-laboratories" one does, which was common when the error state was previously displayed, but not now.

## Impact
This restores expected behaviour.

## Checklist*


willholt-tb (approved):
must have missed this PR a while ago 😁

PR #526, opened 2025-02-03
fix: skip token check if not using nf tower

PurcyTwoBulls:
## Title*
Skip NF Access Token check if not using NF Tower

## Type of Change*
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
On updates, the api endpoint will check for the NF Access Token, but if the lab is AWS Omics only, it will not have an entry in the parameter store at all for the lab and the get parameter call will fail.

This change will restrict the update call to only check the token if the lab is NF Tower enabled.

## Testing*
Create a lab without seqera
load lab again, don't touch the seqera settings, update an unrelated field (description) and try to save
Will present this error

## Impact
Prevents updating of AWS Omics only labs

## Additional Information
Will now also catch the error if the parameter doesn't exist and send back a different error message. This will make future debugging of errors easier.

## Checklist*


zigtan (approved):
Thanks @PurcyTwoBulls

PR #543, opened 2025-02-11
fix(EG-931): file upload improvements

simondib:
## Title*
Various fixes and improvements for file uploads 

## Type of Change*
- [ ] New feature
- [X] Bug fix
- [ ] Documentation update
- [X] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [X] UI/UX improvement

## Description
- Reverted file duplication check to previous implementation for robustness
- Modified `getFileDetails` to properly reset file state - this fixes a bug when trying to re-add a file that had been prevoip:
    - `progress: undefined`
    - `error: undefined`
    - `url: undefined`

- Improved validation logic for file pairs
- Refactored boolean check for s3 url availability and re-wired the `Download sample sheet` button so it displays at the same time as the sample sheet URL bar

## Testing*
Manually tested scenarios

## Checklist*

### DIFF
--- a/packages/front-end/src/app/components/EGRunFormUploadData.vue
+++ b/packages/front-end/src/app/components/EGRunFormUploadData.vue
@@ -302,8 +302,8 @@
     return fileName.substring(0, fileName.lastIndexOf('_R'));
   }
 
-  function toggleDropzoneActive() {
-    isDropzoneActive.value = !isDropzoneActive.value;
+  function setDropzoneActive(val: boolean) {
+    isDropzoneActive.value = val;
   }
 
   async function startUploadProcess() {
@@ -667,10 +667,10 @@
     >
       <div
         id="dropzone"
-        @dragenter.prevent="toggleDropzoneActive"
-        @dragleave.prevent="toggleDropzoneActive"
+        @dragenter.prevent="setDropzoneActive(true)"
+        @dragleave.prevent="setDropzoneActive(false)"
         @dragover.prevent
-        @drop.prevent="toggleDropzoneActive"
+        @drop.prevent="setDropzoneActive(false)"
       >
         <div
           :class="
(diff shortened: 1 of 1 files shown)


packages/front-end/src/app/components/EGRunFormUploadData.vue
```diff
@@ -56,7 +57,16 @@
     wipRunTempId: string;
   }>();
 
+  const MAX_FILE_SIZE = 5 * 1024 * 1024 * 1024; // 5GB
+  const MIN_FILE_SIZE = 1; // 1byte
+  const UPLOAD_TIMEOUT = 600000; // 10 mins
+  const UPLOAD_RETRY_DELAY = 3000; // 3 second delay helps prevent immediate retry spam and gives time for temporary network issues to resolve
+
   const chooseFilesButton = ref<HTMLButtonElement | null>(null);
+  const isDropzoneActive = ref(false);
```
willholt-tb, line 66: re: this ref, the function `toggleDropzoneActive`, and this markdown
```
      <div
        id="dropzone"
        @dragenter.prevent="toggleDropzoneActive"
        @dragleave.prevent="toggleDropzoneActive"
        @dragover.prevent
        @drop.prevent="toggleDropzoneActive"
      >
```

the three events which invoke a toggle:

`dragenter`, `dragleave`, `drop`

is it completely consistent which state the dropzone should be in after each of those?

If so, I think it would be safer to set the value explicitly rather than toggle, eg.

```
        @dragenter.prevent="setDropzoneActive(true)"
        @dragleave.prevent="setDropzoneActive(false)"
        @dragover.prevent
        @drop.prevent="setDropzoneActive(false)"
```

which prevents any issues where the state of the ref gets inverted somehow, and the toggles end up doing the opposite of what they are supposed to do.

willholt-tb (approved):

PR #544, opened 2025-02-11
feat(EG-995): bundle icon set in client build

simondib:
## Title*
Bundle icon set in client build 

## Type of Change*
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [X] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Previously, the app relied on Iconify's CDN to serve icons, which caused issues when users were offline or experienced connectivity problems. This wasn't ideal since icons would fail to load when the app needed to render client-side code without an internet connection.

The solution moves all icons into a local bundle that gets built with the client code. This removes dependency on external services to display icons, making the app more reliable and self-contained. It's a simpler setup that works consistently whether online or offline, and the icons load faster since they're served directly from the build.

This change makes the icon system more robust and provides better control over assets, while also improving the user experience during offline scenarios.

## Testing*
All icon instances were visually verified on a deployed environment.

## Impact
This was mostly a configuration change with minimal template change. 

## Checklist*


zigtan (approved):
Thanks @simondib

PR #546, opened 2025-02-13
feat(EG-1004): update the create-file-upload-sample-sheet API to support single read files

zigtan:
## Update create-file-upload-sample-sheet API to support single-read / paired-read files
This PR updates the `/easy-genomics/upload/create-file-upload-sample-sheet` API to support single-read and paired-read sequenced files to generate the sample-sheet CSV file.

It also adds data validation and improved error messaging for the FE to handle as necessary.

## Type of Change*
- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [X] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
This API will now handle requests that define:
 - one or more single R1 files
 - one or more paired R1 and R2 files

All other scenarios will return an InvalidRequest (HTTP 400) error with the appropriate message.

The sample-sheet generation also checks if the R1 / R2 file starts with the supplied `SampleId` value, and if it doesn't it will also return an InvalidRequest (HTTP 400) error.

The API's response has also been enhanced to include the `SampleSheetType` property that will either have: a `single-read` or a `paired-read` value.

## Testing*
See associated JIRA ticket for testing instructions.

## Impact
None.

## Additional Information
This API is dependent on the FE supplying the `SampleId` and it uses this value to check if the R1 / R2 file starts with the `SampleId`. 

## Checklist*

### DIFF
--- a/packages/back-end/src/app/controllers/easy-genomics/upload/create-file-upload-sample-sheet.lambda.ts
+++ b/packages/back-end/src/app/controllers/easy-genomics/upload/create-file-upload-sample-sheet.lambda.ts
@@ -41,7 +41,7 @@ export const handler: Handler = async (
     const requestParseResult = SampleSheetRequestSchema.safeParse(request);
     if (!requestParseResult.success) {
       requestParseResult.error.issues.forEach((issue) => console.error(issue));
-      throw new InvalidRequestError();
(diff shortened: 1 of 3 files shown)


willholt-tb (approved):
thanks Zig!

PR #560, opened 2025-02-20
feat(EG-978): update Step 3 UI to display & validate required parameters

zigtan:
## Title*
Update Step 3 UI to display & validate required parameters for Seqera Pipelines and AWS HealthOmics Workflows

## Type of Change*
- [X] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
This PR adds the display of `*` along side the name of required parameters for Seqera Pipelines and AWS HealthOmics Workflows, and adds validation logic to check the required parameters are not empty. If there are empty required parameters it will display an error dialog message indicating which need to be updated.

This PR has been updated to also fix some existing bugs with the Seqera Pipeline Step 3 form behaviour:
* Updating the `input` field no longer changes the generated sample-sheet url value.
* Pass through the updated values to Step 4, and retain the updated values if switching back to Step 3.

## Testing*
Test launching a Seqera Pipeline or AWS HealthOmics Workflow using the Easy Genomics stepper UI.

## Impact
None.

## Additional Information
None.

## Checklist*

### DIFF
--- a/packages/front-end/src/app/pages/labs/[labId]/run-pipeline/[pipelineId].vue
+++ b/packages/front-end/src/app/pages/labs/[labId]/run-pipeline/[pipelineId].vue
@@ -94,12 +94,20 @@
       .filter((_) => _)
       .reduce((acc, cur) => ({ ...acc, [Object.keys(cur)[0]]: Object.values(cur)[0] }), {});
 
+    // Identify Seqera pipeline schema required parameters
+    const paramsRequired: string[] = definitions.input_output_options.required
+      ? definitions.input_output_options.required
+      : [];
+
     schema.value = {
       ...originalSchema,
       $defs: filteredDefinitions,
     };
     if (pipelineSchemaResponse.params) {
-      runStore.updateWipSeqeraRun(seqeraRunTempId.value, { params: JSON.parse(pipelineSchemaResponse.params) });
+      runStore.updateWipSeqeraRun(seqeraRunTempId.value, {
+        params: JSON.parse(pipelineSchemaResponse.params),
+        paramsRequired: paramsRequired,
+      });
     }
   }
 
--- a/packages/front-end/src/app/pages/labs/[labId]/run-workflow/[workflowId].vue
+++ b/packages/front-end/src/app/pages/labs/[labId]/run-workflow/[workflowId].vue
@@ -45,8 +45,7 @@
 
   const loading = computed<boolean>(() => uiStore.isRequestPending('loadOmicsWorkflow'));
 
-  onBeforeMount(loadWorkflow);
-  onMounted(initializeWorkflowData);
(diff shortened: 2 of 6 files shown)


willholt-tb:
@zigtan what do you think about wrapping the parameter fields in a UForm so that we get this kind of validation on them instead of doing it manually in toasts?
image

packages/front-end/src/app/pages/labs/[labId]/run-pipeline/[pipelineId].vue
```diff
@@ -94,12 +94,20 @@
...
+    // Identify Seqera pipeline schema required parameters
+    const paramsRequired: string[] = definitions.input_output_options.required
+      ? definitions.input_output_options.required
+      : [];
+
     schema.value = {
       ...originalSchema,
       $defs: filteredDefinitions,
     };
     if (pipelineSchemaResponse.params) {
```
Comment (willholt-tb) [inline packages/front-end/src/app/pages/labs/[labId]/run-pipeline/[pipelineId].vue:106]: I don't think the setting `paramsRequired` should be inside this conditional. In fact, does the conditional need to exist at all? Could we just have eg.

```
runStore.updateWipSeqeraRun(seqeraRunTempId.value, {
  params: JSON.parse(pipelineSchemaResponse.params || '{}'),
  paramsRequired: paramsRequired,
});
```

packages/front-end/src/app/pages/labs/[labId]/run-workflow/[workflowId].vue
```diff
@@ -96,10 +84,26 @@
...
+          return paramName;
+        }
+      })
+      .filter((_) => _ != undefined);
+
     // initialize params if they aren't present already
     if (!runStore.wipOmicsRuns[omicsRunTempId.value].params) {
       runStore.updateWipOmicsRun(omicsRunTempId.value, {
         params: {},
+        paramsRequired: paramsRequired,
```
Comment (willholt-tb) [inline packages/front-end/src/app/pages/labs/[labId]/run-workflow/[workflowId].vue:106]: I think this should also not be behind the `params` conditional

willholt-tb (approved):

packages/front-end/src/app/pages/labs/[labId]/run-pipeline/[pipelineId].vue
Comment (zigtan) [reply packages/front-end/src/app/pages/labs/[labId]/run-pipeline/[pipelineId].vue:106]: It looks like the IF statement was already existing to enforce params exists since it is an optional property. 

But I agree it can be simplified to your suggested implementation.

packages/front-end/src/app/pages/labs/[labId]/run-workflow/[workflowId].vue
Comment (zigtan) [reply packages/front-end/src/app/pages/labs/[labId]/run-workflow/[workflowId].vue:106]: @willholt-tb , I'll create a separate ticket to address this change.

PR #577, opened 2025-03-05
feat(EG-1035): reset step 3 params to defaults on all file removal

simondib:
## Title*
Resets Stepper step 3 params to default values when all file uploads are removed.

## Type of Change*
- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
[Provide a detailed explanation of the changes you have made. Include the reasons behind these changes and any relevant context. Link any related issues.]

## Testing*
Manually tested multiple files uploads and removal, then verified params matched default values.

## Checklist*


willholt-tb:
Re-running initializePipelineData will also clear the run name entered in step 1. It's a hard reset of the process. I think we need to kick the user back to the start with everything reset.

My forthcoming PR to fix the Launch Another Workflow button will actually do this, I've added the initial loading state that the workflow page has to the pipeline page too (side note, it was a placeholder and we should think about a nicer looking design). 

My only concern is that it might be annoying to the user when they clear the last file to get kicked back to the start, but I think that's what functionally needs to happen.

willholt-tb:
I notice this is only applied to Seqera pipelines, should the same behaviour be applied to Omics workflows?

simondib:
Closed as addressed in https://github.com/twobulls/easy-genomics/pull/582

PR #595, opened 2025-03-11
feat(EG-1055): open file explorer to /results

willholt-tb:
## Title*

Open File Browser to `/results`

## Type of Change*
- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description

When the File Browser loads in the S3 tree info, if the root dir contains a `results` dir in its children, it takes the step into that dir.

## Impact

This will happen anywhere we have a File Browser - right now I believe that's just the results page but if we use it elsewhere we might need to make this feature optional.

## Checklist*

### DIFF
--- a/packages/front-end/src/app/components/EGFileExplorer.vue
+++ b/packages/front-end/src/app/components/EGFileExplorer.vue
@@ -26,6 +26,7 @@
       s3Prefix: string;
       s3Contents: S3Response | null;
       isLoading?: boolean;
+      startPath?: string[];
     }>(),
     { isLoading: true },
   );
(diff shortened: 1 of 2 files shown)


zigtan:
@willholt-tb , we need to leverage the LaboratoryRun object's `InputS3Url` and `OutputS3Url` (aka `"results"` directory) which the User can modify in Step 3 to another path name if the choose instead of the default `"results"` path name.

willholt-tb:
@zigtan understood, will do 👍

willholt-tb:
@zigtan it works!
image

willholt-tb:
@zigtan the omics run does generate a sub dir with the id you were talking about which isn't included in the outdir variable though
image

zigtan:
Nice work @willholt-tb ! Yeap thanks for reminding me that Omics adds their unique id reference after the processing has completed. I think with your implementation in this PR it meets our end-user's needs already.

I believe the unique id reference is the same external id reference returned after launching the Omics workflow, so we can add some logic in Easy Genomics to bridge this S3 path.

zigtan (approved):
Great work @willholt-tb

PR #609, opened 2025-03-19
feat(EG-1052): meta PR

willholt-tb:
Changes:

- bc8e3ac: refactor(EG-1052): rename currentLab -> mostRecentLab because unlike currentOrg there is not a global current lab

Thinking in terms of "current lab" is subtly very misleading. This has previously caused components (specifically the pinia stores) to be designed under the assumption of there being a single, current lab, like an app. This has incurred sizeable refactoring work in the past because EasyG is a website and can have multiple tabs with multiple labs open at the same time (unlike orgs - there really is a single current org). I've called the lab store `mostRecentLab` because that's what it really is.

- 9758fba: refactor(EG-1052): move most recent lab logic to root page

The "open the most recent lab if available" logic is useful outside of just signing in. A user might have an EasyG bookmark that we want to open to their last used lab whatever it may be. To enable this, I moved the direction logic to the root path `/`, instead of in the sign in procedure, and direct to `/` post-sign in so that continues to work the same way.

- 8f5f95e: refactor(EG-1052): use null instead of empty string for unset most recent lab

If a value is not present, I tend to use `null` as a more obvious representation of this than `''`. This pattern is in use throughout the FE so this is consistent with established patterns.

- ca85d71: refactor(EG-1052): set mostRecentLab on lab view

I made the most recent lab update as simple and consistent as possible by putting it in the `onMounted` of the lab view.

I also tweaked the "load default org and lab" logic to not overwrite an existing most recent lab value, as that function gets called on every page view. I commented this better in 25ce670.

- 6c34f95: refactor(EG-1052): unsuppress error if user id is missing which should never be the case

I removed a conditional which should be an invalid case and surfaced the error instead, so that we don't miss that something is going wrong if it does somehow happen

- 2d399cd: refactor(EG-1052): read orgId from lab to prevent potential data conflict

The lab that `LabView` reads from is specified by `props.labId` and includes the org id on the lab object. Patching in the org id from a different source (the user store) introduces the possibility of a data conflict if our approach to organization handling ever changes, or the `LabView` component is ever used to display a lab cross-organization. It shouldn't matter now, but it could save some headaches in the future.

- d8bd253: refactor(EG-1052): load all labs if required from breadcrumb so lab page can only load itself

I offloaded the responsibility of fetching all labs if not present to the labs breadcrumb, so that the LabView always just needs to worry about loading its own lab.

### DIFF
--- a/packages/front-end/src/app/components/EGBreadcrumbLabs.vue
+++ b/packages/front-end/src/app/components/EGBreadcrumbLabs.vue
@@ -11,12 +11,26 @@
 
   const isOpen = ref<boolean>(false);
 
+  const currentOrgId = computed<string | null>(
+    () => (userStore.isSuperuser ? (route.params.orgId as string) : userStore.currentOrgId) ?? null,
+  );
+
+  onBeforeMount(async () => {
+    // there are some cases (eg. sign in straight to a lab page) where the labs might not have been loaded
+    // in that case we'll fetch them now
+
+    if (currentOrgId.value === null) return; // this is probably a problem but nothing we can do
+
+    if (!(currentOrgId.value in labsStore.labIdsByOrg)) {
+      await labsStore.loadLabsForOrg(currentOrgId.value);
+    }
+  });
+
   const currentLab = computed<Laboratory | null>(() => labsStore.labs[labId] || null);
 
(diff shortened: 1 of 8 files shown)


zigtan (approved):
LGTM.

PR #620, opened 2025-03-26
feat(EG-1054): add filtering support for list-laboratory-runs API

zigtan:
## Title*
Add filtering support for list-laboratory-runs API

## Type of Change*
- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
This PR adds logic to support filtering for the `/easy-genomics/laboratory/run/list-laboratory-runs` API to support filtering by one or more properties in the `LaboratoryRun` object type through the use of query parameters.

The filtering is performed in the order of the supplied query parameters, and the query parameters must match the `LaboratoryRun`'s property names and property value stored in the database record (i.e. it is case-sensitive).

e.g.
 - `/easy-genomics/laboratory/run/list-laboratory-runs?LaboratoryId={LabId}&Status=COMPLETED`

The filtering also supports partial filtering by the supplied value with the use of '*' character.

e.g.
 - `/easy-genomics/laboratory/run/list-laboratory-runs?LaboratoryId={LabId}&RunName=*-my-test-run-name` // filter records with RunName ending with `-my-test-run-name`
 - `/easy-genomics/laboratory/run/list-laboratory-runs?LaboratoryId={LabId}&RunName=my-test-run-name-*` // filter records with RunName starting with `my-test-run-name-`

The filtering logic is reusable and can be easily added to other List APIs to provide the same filtering behavior with respect to its object's own properties.

## Testing*
See associated JIRA ticket for testing instructions.

## Impact
None.

## Additional Information
None.

## Checklist*

### DIFF
--- a/packages/back-end/src/app/utils/rest-api-utils.ts
+++ b/packages/back-end/src/app/utils/rest-api-utils.ts
@@ -125,3 +125,29 @@ export function getAwsHealthOmicsApiQueryParameters(event: APIGatewayProxyEvent)
 
   return apiQueryParameters;
 }
+
+/**
+ * Helper function to return valid query parameters that exist in the object's properties.
+ * @param objectProperties
+ * @param queryParameters
+ */
+export function getFilters(objectProperties: string[], queryParameters: [string, string][]): [string, string][] {
+  return queryParameters.filter((_: [string, string]) => objectProperties.includes(_[0]));
+}
+
+/**
+ * Helper recursive function to apply filters object records by the supplied filters.
+ * @param results
+ * @param filters
+ */
+export function getFilterResults<T>(results: T[], filters: [string, string][]): T[] {
+  if (filters.length === 0) {
+    return results;
+  }
+
+  const filter: [string, string] = filters.shift();
+  const filterResults = results.filter((r: T) => r[filter[0]] === filter[1]);
+
+  // Recursion
+  return getFilterResults<T>(filterResults, filters);
+}
--- a/packages/back-end/src/app/utils/rest-api-utils.ts
+++ b/packages/back-end/src/app/utils/rest-api-utils.ts
@@ -145,8 +145,16 @@ export function getFilterResults<T>(results: T[], filters: [string, string][]): 
     return results;
   }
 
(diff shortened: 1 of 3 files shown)


packages/back-end/src/app/utils/rest-api-utils.ts
```diff
@@ -125,3 +125,37 @@ export function getAwsHealthOmicsApiQueryParameters(event: APIGatewayProxyEvent)
...
+
+  const [filterKey, filterVal] = filters.shift();
+  const filterResults = results.filter((r: T) => {
+    if (filterVal.startsWith('*')) {
+      return r[filterKey].endsWith(filterVal.slice(1)); // Ignore leading *
+    } else if (filterVal.endsWith('*')) {
+      return r[filterKey].startsWith(filterVal.slice(0, -1)); // Ignore trailing *
+    } else {
+      return r[filterKey] === filterVal;
+    }
```
willholt-tb, line 156: This handling of the wildcards feels a bit limiting, cause you can only have them on the front/end and you can only have one or the other. As a user I would like to be able to use `*` anywhere in the string.

I don't have a full working alternative as I write this though - my thought process:

- use a regex and turn `*` into `.*`
  -  but the rest of the user input needs to be handled as a string literal, not regex
- escape the rest of the user string
  - but there doesn't seem to be a widely available regex escape function https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape 
    - but there are other options that seem feasible https://stackoverflow.com/a/6969486
- or we could write an algorithm to do the matching logic manually

Given that I'm the one complaining I'd be happy to tackle implementing this. I'm also aware that this could be performance sensitive. 

@zigtan if you're happy with this I'll have a crack at making it work.

zigtan: Nice suggestion @willholt-tb . Yep, I'm happy for you to try to tackle this enhancement to the filtering behavior in a separate ticket/PR.

willholt-tb (approved):

PR #632, opened 2025-04-01
fix(EG-1096): switch org problems

willholt-tb:
## Title*

Fix problems with org switching:
- unexpected delay on switch
- current lab persisted across org switch

## Type of Change*
- [ ] New feature
- [X] Bug fix
- [ ] Documentation update
- [X] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description

Firstly, refactored the org switching code into a component, deduplicating it across header/org breadcrumb.

Then, fixed unexpected delay after org switch visible loading:
- previously, it had the artificial delay, then the network call to update last org (this was the unexpected delay)
- now, the network call is the delay

Also, fixed the current lab persisting across org switch - it was related to the app remount (`incrementRemountAppKey`) interrupting the re-routing logic that would take the page to the proper destination. I fixed it with a 100ms delay; I would much rather remove the app remount, but I don't know why it's there and I don't have the bandwidth atm to figure out if removing it breaks anything.

## Testing*

Tested
- switch between orgs from lab and lab list; last lab feature not implemented for this case
- log out and in from lab and lab list; last lab feature works

## Checklist*

### DIFF
--- a/packages/front-end/src/app/components/EGBreadcrumbOrgs.vue
+++ b/packages/front-end/src/app/components/EGBreadcrumbOrgs.vue
@@ -1,11 +1,8 @@
 <script setup lang="ts">
   import { Organization } from '@easy-genomics/shared-lib/src/app/types/easy-genomics/organization';
-  import { ButtonVariantEnum } from '@FE/types/buttons';
 
-  const router = useRouter();
   const route = useRoute();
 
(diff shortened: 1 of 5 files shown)


zigtan (approved):
Thanks @willholt-tb .

PR #634, opened 2025-07-10
feat: run status update redesign

k-florek:
## Type of Change
- [X] New feature
- [X] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
This PR includes a new lambda function that handles requests from the front-end to check the status of any run that is not in a terminal state. The lambda function publishes an message to the laboratory run update topic which eventually causes the process-update-laboratory-run to determine the status of the runs. A change was added to the front-end where a table refresh key was added and the fetchLaboratoryRuns() function was removed. This enables the table to be updated and refreshed without the user needing to manually reload the page.

I've also included an unrelated minor bug fix that removes the spaces from samplesheet, which causes problems in a number of pipelines.

## Testing
I have tested this with healthomics runs to see the status update changes. I am unsure if any built in validations or tests need to be changed.

## Impact
This change solves an issue where runs that take hours - days to complete would have incorrect statuses because AWS would halt the process-update-laboratory-run lambda function due to a recursion limit. This update circumvents this by removing the recursion from process-update-laboratory-run and moving the logic to the front-end to periodically check for status updates.

## Additional Information
I believe I have added this new functionality to be inline with the project design, however it is likely I overlooked something in this implementation. So please review with care.

## Checklist

### DIFF
--- a/packages/back-end/src/app/controllers/easy-genomics/laboratory/run/process-update-laboratory-run.lambda.ts
+++ b/packages/back-end/src/app/controllers/easy-genomics/laboratory/run/process-update-laboratory-run.lambda.ts
@@ -15,14 +15,14 @@ import { SQSEvent } from 'aws-lambda/trigger/sqs';
 import { LaboratoryRunService } from '@BE/services/easy-genomics/laboratory-run-service';
 import { LaboratoryService } from '@BE/services/easy-genomics/laboratory-service';
 import { OmicsService } from '@BE/services/omics-service';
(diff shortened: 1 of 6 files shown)


zigtan (approved):
Thanks @k-florek ! Nice work on adding the new API and fixing the sample sheet format.

If status update check hits performance limits with Seqera's APIs we might need to revisit the solution.

PR #681, opened 2026-03-18
feat: omics run tag based access controls

kevin-loftus-dept:
## Type of Change*
- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [X] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Implemented lab-scoped IAM-enforced access control for AWS Omics runs via run tagging.
- Added Omics access role + IAM policies to enforce
    - `omics:StartRun` allowed only when `aws:RequestTag/LaboratoryId` and `aws:RequestTag/OrganizationId` match the caller’s `aws:PrincipalTag/*`.
    - `omics:GetRun`, `omics:ListRuns`, and `omics:CancelRun` allowed only for Runs tagged with the matching `aws:ResourceTag/LaboratoryId`.
- Refactored existing Omics calls to use lab-scoped credentials
    - Added a reusable STS AssumeRole helper (`packages/back-end/src/app/services/sts-service.ts`) that sets session tags (`LaboratoryId`, `OrganizationId`, optional `UserId`).
- Added a backfill script (`packages/back-end/scripts/backfill-omics-run-tags.ts`) that scans `laboratory-run-table` and applies as many corresponding tags as possible to the pre-existing data.

## Testing*
Unit tests, ran the app locally.

## Impact
Users can now only list / read / cancel Omics Runs that are tagged with their laboratory’s LaboratoryId, and create new Runs only when StartRun request tags (LaboratoryId, OrganizationId) match the lab-scoped IAM session.

Operationally, the backend now performs Omics calls using STS-assumed, lab-tagged credentials, removing reliance on broad Omics permissions shared across all users. 

Security-wise, this reduces the risk of cross-laboratory data exposure by enforcing the lab boundary directly at the AWS Omics API permission layer rather than only via application-side checks.

**Note**: When we deploy to higher environments, we will need to remember to run the one-off script that populates run tags on existing data.

## Checklist*


packages/back-end/package.json
```diff
@@ -25,7 +25,9 @@
     "lint": "eslint 'src/**/*.{js,ts}' --fix",
     "local-server": "tsx src/local-server/index.ts",
     "local-server:watch": "tsx watch src/local-server/index.ts",
-    "invoke-process-handler": "tsx src/local-server/invoke-process-handler.ts"
+    "invoke-process-handler": "tsx src/local-server/invoke-process-handler.ts",
```
ignacionistal, line 28: @kevin-loftus-dept the scripts need to be added to the .projenrc.ts, after that you can run projen command and package.json will be re generated.

kevin-loftus-dept: Thanks for calling that out, I have added the scripts into the root `.projenrc.ts` file.

packages/back-end/src/infra/stacks/aws-healthomics-nested-stack.ts
```diff
@@ -144,6 +152,40 @@ export class AwsHealthOmicsNestedStack extends NestedStack {
...
+    this.iam.addPolicyStatements('omics-access-role-permissions-policy-statement', [
+      // Allow StartRun only when request tags match principal tags for lab and org
+      new PolicyStatement({
+        resources: ['*'],
+        actions: ['omics:StartRun'],
+        effect: Effect.ALLOW,
+        conditions: {
+          StringEquals: {
+            'aws:RequestTag/LaboratoryId': '${aws:PrincipalTag/LaboratoryId}',
+            'aws:RequestTag/OrganizationId': '${aws:PrincipalTag/OrganizationId}',
```
kevin-loftus-dept, line 167: Do we need org ID as well?

juanarroqui (approved):

packages/back-end/src/infra/stacks/aws-healthomics-nested-stack.ts
juanarroqui: probably lab id is enough, but not really sure, may use both just to be sure. Unless you can find some other parts of the app where they use lab id only

PR #686, opened 2026-03-23
feat: server-side pagination for runs table

kevin-loftus-dept:
## Type of Change*
- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [X] UI/UX improvement

## Description

Implements server-side cursor pagination for the Pipeline Runs table on the lab view (EGLabView), replacing client-side slicing of the full runs list so the browser no longer loads hundreds of rows at once.

### Backend

- Extends GET /laboratory/run/list-laboratory-runs with a serverMode=true path that returns { items, hasMore, nextToken }.
- Uses DynamoDB Query on the lab partition with CreatedAt_Index / ModifiedAt_Index for sort, Limit, and ExclusiveStartKey (cursor encoded in nextToken).
- Supports UserId (e.g. “My runs only”) and structured search for status only ("status"=...).
- In local/dev, optionally merges deterministic mock runs and paginates in-memory so pagination can be tested without launching real Omics runs. Non-dev continues to use Dynamo pagination.

### Frontend

- Adds listLabRunsPaginated on the labs repository and wires the Runs tab to fetch pages with sort, filters, and cursor stack for Previous / Next.
- Extends EGTable with paginationMode="server" (Prev/Next) while keeping default client pagination for other tables.
- Updates search helper text to reflect status-only search; “My runs only” still filters by `UserId`, not `Owner`.
- Reduces default page size to 10 rows per request.

## Testing*

- Manual: Pipeline Runs tab with local server — verify Next/Previous, sort (CreatedAt / Last Updated), “My runs only”, and "status"=... search; confirm non-status search is rejected or ignored per current UX.
- Automated:
pnpm --prefix packages/back-end exec jest test/app/controllers/easy-genomics/laboratory/run/list-laboratory-runs.lambda.test.ts --runInBand
pnpm --prefix packages/front-end exec jest test/app/repository/modules/labs.test.ts --runInBand
- Regression: Other EGTable usages remain on client pagination unless pagination-mode="server" is set.

## Impact

- Performance: Smaller payloads and less work in the browser for large labs; Dynamo still does bounded reads per page in prod (not full-table fetch for the paginated path).
- API contract: Legacy callers that omit serverMode still receive the array response; the UI now uses the paginated object when serverMode=true.
- Behavior: No exact total page count; navigation is cursor-based (Previous/Next). Status polling still uses runs visible on the current page (acceptable per current product decision).
- Dev-only: Mock runs in dev/local can inflate list size for QA; prod behavior unchanged aside from the new server-mode endpoint usage from the UI.

## Additional Information

- Sortable columns on the Runs table are limited to what the backend supports (CreatedAt / ModifiedAt via LSIs); other columns are not server-sorted in this iteration.
- Search in server mode is intentionally `status`-only via structured query; use “My runs only” for user-scoped filtering (UserId).
- If we want component tests for EGTable server mode or EGLabView, we can add them in a follow-up (would likely need Vue test utils / similar).

## Screenshots

https://github.com/user-attachments/assets/0d26a212-d2c2-40c5-b415-3476401ce26f

## Checklist*

### DIFF
--- a/packages/back-end/src/app/controllers/easy-genomics/laboratory/run/list-laboratory-runs.lambda.ts
+++ b/packages/back-end/src/app/controllers/easy-genomics/laboratory/run/list-laboratory-runs.lambda.ts
@@ -1,5 +1,9 @@
+import { AttributeValue } from '@aws-sdk/client-dynamodb';
 import { LaboratoryRunSchema } from '@easy-genomics/shared-lib/lib/app/schema/easy-genomics/laboratory-run';
-import { ReadLaboratoryRun } from '@easy-genomics/shared-lib/src/app/schema/easy-genomics/laboratory-run';
+import {
+  ListLaboratoryRunsPaginatedResponse,
+  ReadLaboratoryRun,
+} from '@easy-genomics/shared-lib/src/app/schema/easy-genomics/laboratory-run';
 import { Laboratory } from '@easy-genomics/shared-lib/src/app/types/easy-genomics/laboratory';
 import { LaboratoryRun } from '@easy-genomics/shared-lib/src/app/types/easy-genomics/laboratory-run';
 import { buildErrorResponse, buildResponse } from '@easy-genomics/shared-lib/src/app/utils/common';
@@ -20,6 +24,12 @@ import { getFilterResults, getFilters } from '@BE/utils/rest-api-utils';
 
 const laboratoryService = new LaboratoryService();
 const laboratoryRunService = new LaboratoryRunService();
+
+const SERVER_PAGE_DEFAULT_LIMIT = 25;
+const SERVER_PAGE_MAX_LIMIT = 100;
+const SERVER_MODE_TRUE = 'true';
+const MOCK_RUNS_COUNT = 150;
+const LOCAL_STAGE = 'local';
 
 export const handler: Handler = async (
   event: APIGatewayProxyWithCognitoAuthorizerEvent,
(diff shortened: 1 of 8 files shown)


packages/back-end/src/app/controllers/easy-genomics/laboratory/run/list-laboratory-runs.lambda.ts
```diff
@@ -49,6 +59,61 @@ export const handler: Handler = async (
...
+    const isServerMode = event.queryStringParameters?.serverMode === SERVER_MODE_TRUE;
+
+    if (isServerMode) {
+      const limit = parseLimit(event.queryStringParameters?.limit);
+      const sortBy = parseSortBy(event.queryStringParameters?.sortBy);
+      const sortDirection = parseSortDirection(event.queryStringParameters?.sortDirection);
+      const filters = parseServerFilters(event.queryStringParameters ?? undefined);
+      const nextToken = event.queryStringParameters?.nextToken;
+
+      if (shouldUseMockRuns(event)) {
```
kevin-loftus-dept, line 71: If in local or dev, we pad this data set with mock data so that we can see the pagination in action. We can remove this at a later date if this is no longer wanted or needed.

ignacionistal (approved):
LGTM!

PR #738, opened 2026-05-07
feat: added functionality to create workflow in easy-genomics app

ignacionistal:
## Type of Change*
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Added new form to create Health Omics workflows in application without needed to use AWS console. This form accepts zip files and GitHub urls as sources for the workflow. Also added new field in laboratory setting to set a new GitHub personal access token to be used to pull the workflow definition.

## Testing*
Updated necessary unit tests. Tested on local and dev environments.

## Impact
This does not impact any existing functionality.

## Checklist*

### DIFF
--- a/packages/back-end/src/app/controllers/aws-healthomics/workflow/create-private-workflow.lambda.ts
+++ b/packages/back-end/src/app/controllers/aws-healthomics/workflow/create-private-workflow.lambda.ts
@@ -1,3 +1,4 @@
+import { randomUUID } from 'crypto';
 import { CreateWorkflowCommandInput } from '@aws-sdk/client-omics';
 import { buildErrorResponse, buildResponse } from '@easy-genomics/shared-lib/lib/app/utils/common';
 import {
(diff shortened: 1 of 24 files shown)


kevin-loftus-dept:
I think there is an inconsistency issue with the Github tokens. Please review and let me know if you agree.

Lab GitHub PAT is saved in SSM per org/lab. `create-private-workflow` correctly uses that parameter to pull private repos and build the workflow zip.

## The issue

**What’s inconsistent:** `process-fetch-workflow-schema` (EventBridge) and the GitHub fallback in read-workflow-schema use a single token from Secrets Manager via `GITHUB_PAT_SECRET_NAME`, **not** the lab’s SSM parameter.

**Why that matters:**

Private repos: Workflow creation can succeed (SSM PAT works), but schema fetch can still fail if the Secrets Manager secret is missing, wrong, or scoped to other repos—so users see a created workflow but no / broken schema for the run UI.
Product story: The new lab setting implies “this PAT is what we use for GitHub,” but the schema path does not use that setting unless ops separately maintains the global secret.
Multi-tenant: One shared secret for all schema fetches is a different trust model than per-lab SSM tokens.

## Proposed fix

- Single source of truth: For GitHub API calls that need the lab’s access, read the PAT from SSM using the same path pattern as create-private-workflow.
- How to know which lab: Use OrganizationId / LaboratoryId on the HealthOmics workflow tags (already set at create). For read-workflow-schema, also align with the laboratoryId query param and optionally verify tags match that lab so we don’t mix tenants.
- IAM: Give process-fetch-workflow-schema and read-workflow-schema ssm:GetParameter (decrypt) on the lab GitHub token path; drop unused Secrets Manager access for those flows once migrated.
- Rollout: Either labs set PAT in the app (SSM filled) or ops migrate the old global secret into per-lab SSM where needed; then deprecate GITHUB_PAT_SECRET_NAME for these Lambdas or keep it only as a short-term fallback with clear docs.

diegomayorga-dept (approved):

kevin-loftus-dept (approved):

PR #742, opened 2026-05-12
fix: duplicated data in dynamoDB tables

ignacionistal:
## Type of Change*
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Data for users organizational access and organization's labs where duplicated in out databases. Now we only use tables organization-user-table and laboratory-user-table.

## Testing*
Tested in local environment and development.

## Impact
This has high impact but should not change any functionality.

## Checklist*


kevin-loftus-dept:
Cursor flagged a possible IAM policy permissions issue for the changes on this PR. I investigated a bit and it seems accurate. Can you double check?

# Step-by-step scenario that surfaces the error
1. This branch is merged and the CDK stack is deployed (or a dev/QA environment is updated) so the new Lambda code is live.
2. A user signs in (or an existing session triggers a new ID/access token), so Cognito invokes the Pre token generation trigger with that user’s attributes (including email).
3. The Lambda runs userService.queryByEmail(...) — that succeeds because user-table is still allowed.
4. The Lambda then runs Promise.all([ organizationUserService.queryByUserId(...), laboratoryUserService.queryByUserId(...) ]).
5. Either Query call hits DynamoDB with an identity whose policy does not include those table (and index) ARNs.
6. DynamoDB returns AccessDeniedException (HTTP 400 with AccessDeniedException in the AWS SDK).
7. The Lambda fails; Cognito does not complete token issuance as intended — users see failed login, refresh errors, or generic Cognito errors depending on client and pool configuration.

So the failure is not “wrong business logic”; it is runtime denial on the two new queries after deploy.

# Proposed resolution
Extend the pre-token Lambda’s IAM policy so it can dynamodb:Query the same resources the code actually uses:

- ${namePrefix}-organization-user-table
- ${namePrefix}-organization-user-table/index/* (needed for UserId_Index)
- ${namePrefix}-laboratory-user-table
- ${namePrefix}-laboratory-user-table/index/*

Keep the existing user-table + index/* entries so queryByEmail continues to work.

Implementation-wise, that means updating `this.iam.addPolicyStatements('/auth/process-pre-token-generation', [...])` in auth-nested-stack.ts to add those four ARNs (or a second PolicyStatement with the same actions: ['dynamodb:Query'] for clarity). No need for PutItem, Scan, etc., unless you later change the Lambda.

After the change: redeploy the auth stack (or full app), then repeat sign-in — steps 3–4 succeed and tokens include OrganizationAccess built from membership.

Optional hardening: extend auth-nested-stack.test.ts (the test that asserts env/IAM wiring for /auth/process-pre-token-generation) so a snapshot or assertion includes the new table ARNs, preventing this class of drift on future refactors.

kevin-loftus-dept (approved):

diegomayorga-dept (approved):

ignacionistal:
This PR is not up to date or valid anymore. Re visit this work after data tagging if still valid.

PR #747, opened 2026-05-12
feat: add new error messages for different flows

diegomayorga-dept:
## Title
Improve frontend error handling: replace silent failures with actionable user feedback

## Type of Change
- [x] Bug fix
- [x] UI/UX improvement

## Description

A systematic audit of the frontend codebase (`error-handling-matrix.csv`, 105 rows) identified 20+ error scenarios that were either completely silent, exposed raw error objects in toasts, or showed inline-only errors users could easily miss. This PR addresses all identified gaps across the run wizard, auth flow, store loading, file upload, and global infrastructure.

**Key changes:**

- **Run wizard launch (critical):** If `createLabRun` fails after the external run was already submitted to Seqera or AWS HealthOmics, the run is now live but previously invisible in the app with no recovery path. The fix captures the external run ID before calling `createLabRun` and surfaces it in the error message: *"Your run was submitted but could not be recorded. Contact support with run ID: {id}."* The wizard also no longer cleans up the wip store on this failure path. The single try/catch over multiple sequential API calls in both review components has been split into per-call blocks with distinct, actionable messages.

- **Global Vue error boundary:** Added `error.vue` and `plugins/error-handler.ts` so unhandled render errors show a recovery UI instead of a white screen. The plugin wires up `app.config.errorHandler` for consistent behavior across component-level errors.

- **Store load actions:** `stores/orgs.ts`, `labs.ts`, and `run.ts` load actions had no error handling — API failures left lists silently empty. All load actions now have try/catch with context-specific toasts.

- **Auth silent failures:** Token fetch failure in `repository/factory.ts` previously logged `console.warn` and continued the request without an auth header. Auth middleware failure redirected to sign-in with no explanation. Both now show toasts before redirecting.

- **Task-level run failures:** Failed runs previously showed only a top-level status badge, forcing users to navigate to Seqera Cloud or the AWS Console. The Run Details tab now fetches and displays task-level breakdown using the existing `WorkflowProgressResponse` endpoint (Seqera) and `GetRun` task data (Omics).

- **Upload step toasts:** Three upload failure paths that previously had no toast now show one: manifest API failure, network loss mid-upload, and sample sheet content validation failure.

- **Remaining silent failures:** Fixed across `EGInviteModule.vue`, `accept-invitation.vue`, `EGRunWorkflowFormEditParameters.vue`, `stores/omicsWorkflows.ts`, `EGDashboard.vue`, `EGFileExplorer.vue`, and the repository validation layer.

**Bugs found and fixed during testing:**
- `uploadFiles()` was using `fileDetails.fileName` but `FileDetails` type only has `name` — caused every file upload failure toast to read *"Upload failed for undefined"*
- `local-server/index.ts` Express error handler had only 3 parameters; Express requires exactly 4 to recognise an error handler, causing `TypeError: res.status is not a function` on any unhandled server error

## Testing

Manual testing performed on `localhost:3000` (Nuxt frontend) + `localhost:3001` (local API server). Full test log with exact steps documented in `packages/front-end/src/app/TESTING.md`.

**Confirmed working:**
- FE-3 Error boundary — temp throw in `app.vue`, confirmed error page renders consistently
- FE-4 Store load errors — blocked `list-organizations` and `list-laboratories` endpoints, confirmed toasts fire
- FE-7 Upload toasts — manifest failure (temp throw) and S3 upload failure (network block) both confirmed; filename bug fix also verified
- FE-9 Accept invitation — navigated to `/accept-invitation?invite=fake-token`, confirmed toast on redirect chain
- FE-9 Dashboard pipeline failure — temp throw in `seqeraPipelinesStore.loadPipelinesForLab`, confirmed toast
- FE-9 Invite send failure — blocked invite endpoint, confirmed toast

**Remaining to test manually** (instructions in TESTING.md):
- FE-1/2 Run wizard per-call error messages — requires Step 4 with code injection
- FE-6 Task-level run display — requires a FAILED run in local environment
- FE-9 Clear defaults failure — requires saved Omics workflow defaults

## Impact

- Users will see actionable error messages instead of white screens, silent empty states, or raw technical error objects
- The `createLabRun` critical failure path now gives users a run ID to contact support, preventing data loss scenarios where runs execute on the platform but are invisible in the app
- No new dependencies introduced
- No breaking changes to existing API contracts or component interfaces
- FE-8 (`validateApiResponse` on Seqera pipeline endpoints) was intentionally reverted — the Zod schemas use strict mode and reject valid API response fields (`WorkflowExternalId`, `InputFileKeys`), causing regressions. TODO comments preserved for a future schema alignment task.

## Additional Information

- Network blocking in Chrome DevTools Request conditions was unreliable for `localhost` URLs in this setup. Code injection (temporary throws) was the primary testing method. TESTING.md documents both approaches and the correct URL patterns for staging/production environments.
- The `error-handler.ts` plugin intentionally does **not** call `showError()` for all Vue errors — doing so converted non-fatal API/validation errors into full-page error screens. The plugin logs all errors; `error.vue` is reserved for Nuxt-level fatal errors that Nuxt routes there automatically.

## Checklist
- [x] No new errors or warnings have been introduced
- [x] All tests pass successfully and new tests added as necessary
- [x] Documentation has been updated accordingly (`TESTING.md` added)
- [x] Code adheres to the coding and style guidelines of the project
- [x] Code has been commented in particularly hard-to-understand areas


packages/front-end/src/app/repository/factory.ts
```diff
@@ -81,7 +81,10 @@ class HttpFactory {
       try {
         token = shouldRefresh ? await this.refreshToken() : await getToken();
       } catch (tokenError) {
-        console.warn(`Failed to get token; reason: ${tokenError}; continuing without Bearer or X-API header`);
+        console.error('Failed to get auth token:', tokenError);
+        useToastStore().error('Your session could not be verified. Please sign in again.');
+        await navigateTo('/signin');
+        throw new Error('Auth token unavailable');
```
kevin-loftus-dept: I think we need this code to continue without redirecting or throwing an error here, since we have a few unprotected endpoints that do not require a token. More details:

HttpFactory token failure now aborts every request (factory.ts): Previously, if getToken() failed, the client logged a warning and sent the request without a Authorization header. That behavior is required for unauthenticated easy-genomics calls that still go through HttpFactory.call(), including confirmUserInviteRequest, confirmForgotPasswordRequest, forgotPasswordRequest, and similar flows where there is no Cognito session yet. The new catch path toasts, navigateTo('/signin'), and throws, so those requests never reach the API. That is a correctness / functional regression (invite acceptance and password-reset flows are the highest risk).

kevin-loftus-dept (changes requested):

packages/front-end/src/app/repository/factory.ts
diegomayorga-dept: checking and fixing the flows that do not need authentication

diegomayorga-dept: reverted the change, so we don't break any flow

kevin-loftus-dept (approved):

PR #774, opened 2026-05-27
fix: redirect to correct laboratory when org changed

ignacionistal:
## Type of Change*
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Fixed redirection to wrong lab when user changes org, added guard to validate lab is part of users selected organization.

## Testing*
Tested manually on local environment

## Impact
This impacts lab page.

## Checklist*


kevin-loftus-dept (approved):

diegomayorga-dept:
## PR Review — #774 (fix: redirect to correct laboratory when org changed)

**Summary:** 2 blockers, 9 major, 6 minor.

Worth noting up front: the backend lambdas (`read-laboratory`, `read-laboratory-run`, `list-*`) already enforce org scope server-side, so this guard is a **UX redirect, not a security boundary** — no IDOR survives it. Nothing here is a security blocker. The real issues are the guard failing open and the rule being implemented twice.

---

### Blockers

**1. `hasRedirectedForOrgMismatch` is never reset on `labId` change — `EGLabView.vue` (~L650)**
The `watch(() => props.labId)` resets `hasRefreshedLabForNullToken` and `lastProcessedLabRef`, but not this flag. Once the component redirects once, the flag stays `true` for the reused instance and every future org-mismatch check is permanently blocked.
→ Reset `hasRedirectedForOrgMismatch.value = false` in the `watch(() => props.labId)` callback alongside the existing resets.

**2. Duplicated business logic — `EGLabView.vue` (~L62)**
This PR adds `useEnsureLabInActiveOrg` for the org-mismatch rule, then re-implements the same rule inline (`isLabInActiveOrg` + flag + three redirect blocks). Four redirect sites, two sources of truth — they will drift.
→ Delete the inline implementation and route `EGLabView` through the same composable/helper the page files use.

---

### Major

**3. Fail-open on load failure — `useEnsureLabInActiveOrg.ts` (~L17)**
The `try/catch` around `labsStore.loadLab(labId)` is dead code — `loadLab` swallows its own errors and never throws. On failure `labs[labId]` stays `undefined`, `labOrgId` is `null`, and `!!labOrgId` short-circuits the comparison. API down → user stays on the cross-org page.
→ Replace the catch-based redirect with a presence check: `if (!labsStore.labs[labId]) { await navigateTo(redirectTo); return; }`

**4. Stale persisted store causes false-positive redirect — `EGLabView.vue` (~L79)**
The store uses `persist: true`. The `lab` computed can resolve to stale data from a previous org before `loadLabData()` finishes; the watcher then compares the old `OrganizationId` against the new `currentOrgId`, sees a mismatch, and redirects the user away from a lab they're legitimately allowed to view.
→ Skip the redirect in the watcher while `loadLabData` is pending (mirror the existing `hasNextFlowTowerAccessToken` null-guard pattern in that watcher).

**5. `onMounted` guard races with `onBeforeMount` — `EGLabView.vue` (~L84)**
`onBeforeMount` already runs the guard after `await loadLabData()`. The duplicate in `onMounted` can overlap, both read the flag as `false`, and call `$router.replace` concurrently.
→ Remove the org-mismatch block from `onMounted`; leave it to its UI setup only.

**6. Bare async function, not a composable — `useEnsureLabInActiveOrg.ts` (~L7)**
Every existing composable (`useAuth`, `useMultiplatform`, …) follows `export function useFoo() { … return { … } }`. This is a one-shot async guard with no returned state — it's a utility, not a composable.
→ Move it to a utils module, or restructure to return a method (`return { ensureLabInActiveOrg }`).

**7. Silently swallowed error — `useEnsureLabInActiveOrg.ts` (~L18)**
The `catch` redirects with no toast and no log, so load failures are undiagnosable.
→ Surface the failure with a toast (and `console.error`) before navigating away.

**8. Mixed navigation APIs**
`EGLabView` uses `$router.replace`, the composable uses Nuxt `navigateTo`. Pick one (prefer `navigateTo`). This largely resolves once the duplicate logic in #2 is consolidated.

**9–11. PR checklist / metadata inaccuracies**
- No linked ticket/issue — add `Closes #…`.
- "All tests pass / new tests added" is checked, but no tests are in the diff — uncheck or justify.
- "Documentation updated" is checked, but no docs changed — uncheck or update test docs.

---

### Minor

**12. Double-negation — `useEnsureLabInActiveOrg.ts` (~L28)**
`!!currentOrgId.value && !!labOrgId` → `currentOrgId.value && labOrgId` (values are already `string | null`).

**13. Missing manual-test docs — `useEnsureLabInActiveOrg.ts`**
Composables aren't unit-tested in this repo, so no unit test is required by convention — but the new error/redirect paths aren't documented anywhere. Capture the manual steps (superuser bypass / load failure / mismatch / no-mismatch).

**14. Fail-open render window — `EGLabView.vue` (~L64) / `useEnsureLabInActiveOrg.ts`**
`isLabInActiveOrg()` returns `true` when org/lab id is missing, so a multi-org user deep-linking to a cross-org lab can transiently see the lab UI before the redirect fires. Backend auth still blocks the data; this is cosmetic. Consider holding render behind a loading state until both ids are present.

> Root-cause note: #14 and #3 are the same flaw — the guard answers a three-state question (in-org / out-of-org / **don't-know-yet**) with a boolean, collapsing "don't know" into "allowed." A guard that can't validate should fail closed, not open. Fixing the symptoms is fine for this PR; the pattern itself is worth not copying.

**15. Thin PR description**
Test plan is just "tested manually" — add reproducible steps and a one-line root cause (what state was stale / what guard was missing).

---

☑️ **Issues found** — recommend addressing blockers #1 and #2 before merge; the rest can be folded in or triaged.

diegomayorga-dept (approved):

PR #787, opened 2026-06-09
feat: improve accessibility navigation in dashboard

ignacionistal:
## Type of Change*
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Small improvements for accessibility compliance in application Dashboard page.

## Testing*
Tested manually on local environment.

## Impact
This impacts dashboard page.

## Checklist*

### DIFF
--- a/packages/front-end/src/app/components/EGDashboard.vue
+++ b/packages/front-end/src/app/components/EGDashboard.vue
@@ -30,6 +30,10 @@
   const searchInputId = 'dashboard-global-search';
   const searchResultsListId = 'dashboard-global-search-results';
   const overviewTimeFilterId = 'dashboard-overview-time-filter';
+  const overviewHeadingId = 'dashboard-overview-heading';
+  const recentRunsHeadingId = 'dashboard-recent-runs-heading';
+  const favouriteWorkflowsHeadingId = 'dashboard-favourite-workflows-heading';
+  const highlightedSearchIndex = ref(-1);
 
   interface SearchResult {
     type: 'run' | 'seqera-pipeline' | 'omics-workflow';
@@ -90,6 +94,22 @@
 
   const showDropdown = computed(() => searchFocused.value && searchQuery.value.trim().length > 0);
 
+  const searchStatusMessage = computed(() => {
+    const q = searchQuery.value.trim();
+    if (!q || !searchFocused.value) return '';
+    if (searchResults.value.length === 0) return `No results for "${q}"`;
+    const noun = searchResults.value.length === 1 ? 'result' : 'results';
+    return `${searchResults.value.length} search ${noun} for "${q}"`;
+  });
+
+  watch([searchQuery, searchResults], () => {
+    highlightedSearchIndex.value = searchResults.value.length > 0 ? 0 : -1;
+  });
+
+  function searchOptionId(index: number): string {
+    return `${searchResultsListId}-option-${index}`;
+  }
+
   function selectSearchResult(result: SearchResult) {
     searchFocused.value = false;
     searchQuery.value = '';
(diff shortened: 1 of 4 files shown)


diegomayorga-dept:
## Major

### Conflicting `EGSidebarNav.vue` rewrite vs. PR #782 (coordination)

Both this PR and #782 (same author, both targeting `development`) rewrite the sidebar tab keyboard handling — with *divergent* implementations:

- **#782:** `tabRefs` array + `setTabRef` + `onKeydown` (Up/Down/Home/End, automatic activation).
- **#787:** `focusTab` / `focusPanel` via `getElementById` + `onTabKeydown` (adds Left/Right, `focusPanel` on click).

Whichever merges second will conflict, and the team needs to decide which approach wins (the two aren't compatible — `getElementById` vs. template refs, and #787 moves focus to the panel on click while #782 doesn't).

**Recommendation:** pull the `EGSidebarNav.vue` change out of one of these PRs and land a single agreed version. Not a code bug, but it'll bite at merge time.

---

## Minor

### `EGTable.vue` — focusable `<tr>` has no actionable role

Rows get `tabindex="0"`, `aria-label`, and Enter/Space handlers, but no `role` (e.g. `role="button"`). A focusable `<tr>` with no role is announced as a plain row; screen-reader users won't be told it's activatable.

**Suggestion:** `:role="rowClickAction ? 'button' : undefined"`. Also verify the resulting tab order isn't overly verbose — every row now enters the tab sequence *and* contains inner controls (action menu, links).

### `EGLabView.vue` — `outline-none focus:outline-none` on a `tabindex="0"` panel

Suppresses the focus ring on a focusable tabpanel. Common when focus is moved programmatically and you don't want a visible ring, but if a keyboard user ever tabs into it directly there's no focus indicator. Acceptable, just intentional-flag.

### `onSearchBlur` 200ms `setTimeout`

Slightly fragile pattern to allow `mousedown` selection, but `@mousedown.prevent` already preserves focus, so it works. Pre-existing-ish; no change needed.

---

## Looks good

- Combobox refactor is textbook ARIA 1.2 — `aria-activedescendant`, `role="option"` on `<li>`, options non-tabbable, polite live-region status (`searchStatusMessage`).
- `highlightedSearchIndex` clamping (`Math.min` / `Math.max`) and reset-on-results-change are correct.
- `:key="i"` → `:key="stat.key"` is a real improvement (stable keys).
- `menu-label` is a valid `EGActionButton` prop — verified.
- `<dl>` / `<dt>` / `<dd>` for stats and `<section aria-labelledby>` with proper `<h2>`s are the right semantics; removing the now-redundant sr-only h2 in `EGLabView` is consistent.

---

## Verdict

☑️ **Issues found** — none block correctness, but resolve the **#782 / #787 sidebar collision** before merging both, and consider `role="button"` on actionable rows.

diegomayorga-dept (dismissed):

kevin-loftus-dept (approved):
One optional follow-up for a future PR (not blocking):

Table row activation pattern

Right now, clickable rows use tabindex="0", keyboard handlers, and role="button" on the <tr>, while action menus/buttons live inside the same row. That works, but it creates a known ARIA tension: interactive controls nested inside something announced as a button.

A cleaner long-term pattern:

Don’t make the whole <tr> a button. Keep row click as a mouse convenience if desired.
Put keyboard/screen-reader activation on a specific cell — e.g. make the run name a link/button with a clear label like “View details for {name}”.
Leave action menus as separate controls in the actions column, so there’s no nested interactivity.

PR #798, opened 2026-06-16
feat: add privacy safe analytics implementation

ignacionistal:
## Type of Change*
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
dds privacy-safe, opt-in upstream usage analytics (PostHog) so the project can understand how Easy Genomics is used across deployments. It's a double opt-in — the institution enables it via analytics.enabled in easy-genomics.yaml, and each user must accept an in-app consent banner — and only anonymous, salted/hashed, bucketed data is sent. Off by default, with a dev escape hatch (allow-dev) for hosted demo environments and per-user opt-out in the profile.

## Testing*
Tested manually on local and dev environments.

## Impact
This does not impact any existing feature, it adds analytics in a safe way that should not break any part of the application.

## Checklist*


kevin-loftus-dept:
2 findings, please take a look.

## Finding 1: Cross-browser opt-out not honored

**Description:** Profile opt-out updates DynamoDB, but other browsers that already have local consent `'granted'` never adopt the server `'denied'` value because sync only runs when local consent is `'unset'`.

**Reproduce:**
1. On Browser A, accept the analytics banner (`consent = granted`).
2. On Browser B, sign in as the same user — consent syncs to `granted`.
3. On Browser B, go to Profile → Privacy and turn analytics off (`consent = denied` in DB).
4. On Browser A, sign in and use the app — analytics still fires (local consent remains `granted`).

**Proposed solution:** In `setCurrentUserDataFromToken()`, always honor server `'denied'` (revocation wins). Only adopt server `'granted'` when local is `'unset'`. On applying `'denied'`, call `analytics.optOut()` so PostHog stops immediately, not just on the next profile visit.

---

## Finding 2: Auth routes can leak PII via PostHog pageleave

**Description:** Docs say analytics is forced off on auth routes, but `isSensitiveRoute()` only hides the consent banner — not event capture. With `capture_pageleave: true`, PostHog sends full `$current_url` (including query params) on navigation away from routes like `/signin?email=...`.

**Reproduce:**
1. Accept the analytics banner and use the app normally.
2. Sign out (consent stays `granted`; SDK stays loaded).
3. Trigger a redirect to `/signin?email=user@example.com` (e.g. failed invite flow in `accept-invitation.vue`).
4. Sign in and navigate away — PostHog emits `$pageleave` with the full URL including the email.

**Proposed solution:** Add `isSensitiveRoute()` to `isForceDisabled()` (or `isActive()`). Set `capture_pageleave: false` and rely on the existing sanitized manual `page()` calls, or add a `sanitize_properties` hook that strips query strings from `$current_url` on every event.

kevin-loftus-dept (dismissed):
Approved, though there is still a possible issue with the localstorage solution for a user's opt out setting:

### Stale device consent after sign-out

**Problem:** `analyticsStore.consent` is persisted to `localStorage` and is intentionally *not* cleared on sign-out (`analyticsStore.reset()` is a no-op). On login, `setCurrentUserDataFromToken()` only handles two cases:

- server `denied` → force local opt-out
- server `granted` + local `unset` → adopt server grant

It does **not** handle: server `unset` + local `granted`.

So if User A accepts analytics and signs out, then User B signs in on the same browser (server `AnalyticsConsent: 'unset'`), User B inherits User A's local `granted` state. `isActive()` stays `true` and events are sent without User B ever seeing the consent banner.

---

### Proposed fix (two layers)

#### 1. Clear device consent on sign-out (primary fix)

Device-level consent should be scoped to the signed-in session, not the browser forever. On sign-out, reset local consent to `unset`. The next user (or the same user on re-login) gets the correct state from the JWT sync.

**`useAuth.ts` — `signOut()`**

`async function signOut() {
  try {
    const analytics = useAnalytics();
    analytics.track('signed_out', {});
    analytics.reset();
    // Clear device consent so the next account on this browser is not opted in
    // by default. Server-side consent is restored on next login via JWT sync.
    useAnalyticsStore().setConsent('unset');
    await Auth.signOut();
    useUserStore().reset();
  } catch (error) {
    // ...
  }
}`

**`analytics.ts` store** — update the `reset()` comment (or wire `reset()` to clear consent and call it from `signOut`):

`reset() {
  this.consent = 'unset';
}`

Same-user re-login still works: if their DynamoDB record has AnalyticsConsent: 'granted', setCurrentUserDataFromToken() will re-adopt it when local is unset.

ignacionistal:
> Approved, though there is still a possible issue with the localstorage solution for a user's opt out setting:
> 
> ### Stale device consent after sign-out
> **Problem:** `analyticsStore.consent` is persisted to `localStorage` and is intentionally _not_ cleared on sign-out (`analyticsStore.reset()` is a no-op). On login, `setCurrentUserDataFromToken()` only handles two cases:
> 
> * server `denied` → force local opt-out
> * server `granted` + local `unset` → adopt server grant
> 
> It does **not** handle: server `unset` + local `granted`.
> 
> So if User A accepts analytics and signs out, then User B signs in on the same browser (server `AnalyticsConsent: 'unset'`), User B inherits User A's local `granted` state. `isActive()` stays `true` and events are sent without User B ever seeing the consent banner.
> 
> ### Proposed fix (two layers)
> #### 1. Clear device consent on sign-out (primary fix)
> Device-level consent should be scoped to the signed-in session, not the browser forever. On sign-out, reset local consent to `unset`. The next user (or the same user on re-login) gets the correct state from the JWT sync.
> 
> **`useAuth.ts` — `signOut()`**
> 
> `async function signOut() { try { const analytics = useAnalytics(); analytics.track('signed_out', {}); analytics.reset(); // Clear device consent so the next account on this browser is not opted in // by default. Server-side consent is restored on next login via JWT sync. useAnalyticsStore().setConsent('unset'); await Auth.signOut(); useUserStore().reset(); } catch (error) { // ... } }`
> 
> **`analytics.ts` store** — update the `reset()` comment (or wire `reset()` to clear consent and call it from `signOut`):
> 
> `reset() { this.consent = 'unset'; }`
> 
> Same-user re-login still works: if their DynamoDB record has AnalyticsConsent: 'granted', setCurrentUserDataFromToken() will re-adopt it when local is unset.

@kevin-loftus-dept thanks for reviewing the PR, I agree with your comment. I just pushed a small change to fix this issue by clearing the consent on sign-out.

kevin-loftus-dept:
## Hostname leak in analytics sanitization
### Problem
PostHog auto-attaches URL properties (notably `$current_url` and `$referrer`) to every event. The PR adds `sanitize_properties` in `useAnalytics.ts` to strip query strings and mask UUID-like path segments — but `sanitizeUrl()` still keeps the full origin:

`return `${url.origin}${sanitizePath(url.pathname)}`;`
So events can include values like `https://genomics.customer.edu/labs/:id/runs` instead of a pathname-only value.

That conflicts with the stated privacy guarantees in the docs, consent banner, and config comments: domain names and hostnames are never sent. It also undercuts the design goal of using an opaque random `deployment_id` — the hostname sent to the central PostHog project can identify which institution/deployment each event came from.

PostHog may also attach other host-related properties (e.g. `$host`, `$referring_domain`) that are not currently sanitized.

### Proposed fix
In `packages/front-end/src/app/composables/useAnalytics.ts`:

Change `sanitizeUrl()` to return pathname only (no `url.origin`), e.g. `sanitizePath(url.pathname)`. Apply the same rule to `$referrer`.

Strip or denylist host-related PostHog properties in `sanitize_properties` (or a `before_send` hook), e.g.:

`$host`
`$referring_domain`
any other URL properties PostHog adds by default that include hostname/domain
Keep manual `page()` as-is — it already sends only a sanitized `$pathname`.

Quick sanity check — with analytics enabled and consent granted, inspect a captured event in the browser network tab or PostHog live events and confirm no hostname/domain appears in any property.

kevin-loftus-dept:
## `nuxt-load-configuration-settings.ts` — build blocker

Found merge artifacts in `packages/front-end/nuxt-load-configuration-settings.ts` that break `pnpm run nuxt-load-settings` (and therefore the front-end `build` script).

### 1. Extra closing braces (lines 199–200)

Two stray `}` were added after the `exportNuxtConfigurationSettings(...)` call in the non-CI `else` branch. esbuild/esrun fails with:
`Expected "finally" but found "}"`


**Fix:** Delete lines 199–200. The `else` block should close with a single `}` before `catch`.

### 2. Duplicate imports (lines 1 and 13)

The PR added imports that already exist later in the file:

- `import * as fs from 'fs'` — also on line 22
- `import { loadConfigurations } from '...'` — also on line 19

These are masked by the syntax error today, but once the braces are fixed, esbuild reports `The symbol "loadConfigurations" has already been declared` (same for `fs`).

**Fix:** Remove the redundant imports on lines 1 and 13. Keep the existing grouped import on lines 17–22.

kevin-loftus-dept (approved):

PR #806, opened 2026-06-18
docs(EGV-177): Deploy API Gateway from the OpenAPI spec via SpecRestApi

diegomayorga-dept:
## Title*
EGV-177: Deploy API Gateway from the OpenAPI spec via SpecRestApi

## Type of Change*
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Replaces the imperative `RestApi` + per-route `addMethod`/`LambdaIntegration` wiring with `SpecRestApi`, sourcing both API Gateways directly from `packages/shared-lib/src/app/openapi/easy-genomics-api.yaml`. The spec becomes the single deployment source of truth, and a spec/route mismatch now fails at **synth** instead of at runtime.

The committed spec stays clean and AWS-neutral; all API Gateway extensions are injected at synth by a new construct:

- **`shared-lib/.../load-spec.ts`** — parses the spec (keeps the `js-yaml` dependency on the shared-lib side of the pnpm boundary).
- **`openapi-spec-enrichment.ts`** (pure / CDK-free, unit-tested) — injects `x-amazon-apigateway-integration` (`aws_proxy`) per operation, converts the `cognitoJwt` bearer scheme into a Cognito User Pools authorizer, adds mock CORS `OPTIONS` per path (replicating the previous `defaultCorsPreflightOptions`), filters paths by prefix per API, and **throws when a spec route has no backing Lambda**. It also downgrades the in-memory document from OpenAPI 3.1 → 3.0.1 (converting `type: [T, null]` unions to `nullable`) because API Gateway's importer only accepts 3.0.x.
- **`spec-rest-api-construct.ts`** — builds the `SpecRestApi` and grants one `lambda:InvokeFunction` permission per backing function.

`LambdaConstruct` now only builds functions, wires `process-*` event sources, and exposes them via `lambdaFunctions`; route registration is gone. Both stacks build their nested stacks first, then construct the API from the aggregated function maps. The obsolete dead `api-gateway-construct.ts` and the unused `restApi` prop were removed.

**Physical-ID preservation:** the new construct reuses the exact ids the old `RestApi` used, so the `AWS::ApiGateway::RestApi` logical id — and therefore the deployed REST API id and invoke URL — is unchanged. This is required because Easy Genomics runs in customer-owned accounts; a replaced REST API would change every customer's API URL.

## Testing*
- `cdk synth` succeeds for the full app; both stacks pass the synth-time resource budget guardrail (the easy-genomics API stack was previously **failing** it at 463).
- Verified against the pre-change baseline template: RestApi logical id `…apigw6039DC28` **preserved** in both stacks; stage `prod` unchanged; synthesized body is OpenAPI 3.0.1 with **0** remaining 3.1 union types; **102** routes total (78 easy-genomics + 11 aws-healthomics + 13 nf-tower).
- New unit tests: `openapi-spec-enrichment.test.ts` (integration/authorizer/CORS injection, prefix filtering, `/{id}` join, 3.1→3.0 downgrade, and the **typo→synth-failure** guard) and `spec-rest-api-construct.test.ts`. All 22 pass.
- Existing suites green: 38 back-end infra tests, `openapi-guard.test.ts` (4).
- **Pending (require deploy to `quality`, not runnable locally):** `cdk diff` against the deployed stack to confirm the RestApi is **modified-in-place, not replaced**; and `pnpm run test-e2e`.

## Impact
- **Resource-limit relief:** the easy-genomics API stack drops from **463 → 96** CloudFormation resources (Method/Resource/per-method permission amplification collapses into the single `RestApi` definition body). This clears the stack off the 500-resource wall it was already tripping. The Lambda nested stacks are unchanged (~388) — this PR does **not** address that separate wall.
- No runtime behavior change intended: same routes, methods, Cognito auth, and CORS. Invoke permissions are created in the API (parent) stack rather than the Lambda nested stacks, to avoid regressing the near-limit nested stacks.
- No new runtime dependencies; `js-yaml` (already a shared-lib dep) is now used at synth.

## Additional Information
- **Deploy-time watch item:** API Gateway must accept the synthesized spec body. The 3.1→3.0 downgrade mitigates the known importer incompatibility, but it can only be confirmed on the first deploy to `quality`.
- Per-method `AwsSolutions-APIG4/COG4` nag suppressions for the public routes were removed (those Method resources no longer exist); the public routes' unauthenticated status now lives in the spec (`security: []`) and is guarded by `openapi-guard.test.ts`. If a `CDK_AUDIT` run surfaces a RestApi-level finding, add a single RestApi-scoped suppression.
- Depends on EGV-171 (spec generation), already merged.

## Checklist*


ignacionistal (dismissed):
LGTM

kevin-loftus-dept (approved):

ignacionistal (approved):

PR #842, opened 2026-07-14
fix: load Swagger UI spec as JSON so the docs route works on the local dev server

diegomayorga-dept:
[## Title*
[EGV-176] Load Swagger UI spec as JSON so the API docs route works on the local dev server

## Type of Change*
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
The public API docs route (`GET /easy-genomics/list-api-docs`, added previously) imported the OpenAPI spec via a `.yaml` text import. esbuild bundles that fine when deployed, but the local dev server runs handlers through `tsx`, which has no YAML loader — so locally it parsed the spec as JavaScript and crashed (`SyntaxError: Unexpected number` on `openapi: 3.1.0`). The route was effectively deploy-only.

**Fix — embed the spec as a generated JSON module** (native to esbuild, `tsx`, and Node alike):
- `generate-openapi.ts` now also emits `easy-genomics-api.json` alongside the existing `easy-genomics-api.yaml` (the yaml remains the human-readable / redocly-linted copy; the json is the machine-readable twin the handler imports).
- `list-api-docs.lambda.ts` imports the `.json` instead of the `.yaml`; the now-unused `.yaml` esbuild loader (in `lambda-construct.ts`) and `yaml-module.d.ts` are removed.
- `swagger-ui.ts` now takes a parsed spec object and **clones** it before overriding `servers` (`{ ...spec, servers }`), since the imported JSON is a shared, cached module object that must not be mutated across warm invocations.

**Two local-only parity fixes** so the route behaves the same locally as when deployed:
- Added `GET /easy-genomics/list-api-docs` to the local dev server's `UNAUTHENTICATED_PATHS` (`auth-middleware.ts`) — it was returning `EG-101 (Authorization header missing)` locally despite being public when deployed (the local server has its own allowlist, separate from the CDK `authorizer: undefined`).
- The handler now falls back to the `Host` header for the "Try it out" base URL when `requestContext.domainName` is absent (local), keeping the `https://<domain>/<stage>` behaviour when deployed.

Out of scope: the committed `easy-genomics-api.yaml` is stale relative to current `ROUTE_SCHEMAS` on `development` (unrelated routes not regenerated by their authors) — deliberately left untouched here to keep this fix focused.

## Testing*
- Unit tests pass locally: `jest src/app/openapi/swagger-ui.test.ts src/app/openapi/openapi-guard.test.ts` → **8/8**. `swagger-ui.test.ts` was updated to the object-based helper API and gains a test asserting the source spec is not mutated.
- ESLint clean on all changed files.
- Verified the generated `easy-genomics-api.json` is valid and resolves as an import under `tsx` (`openapi 3.1.0`, 105 paths, `list-api-docs` present) — i.e. the original local crash no longer reproduces at import time.
- **Not yet verified (please confirm before merge):**
  - Local render — restart the dev server (`pnpm --prefix packages/back-end run local-server`) and open `http://localhost:3001/easy-genomics/list-api-docs`; expect the page with no auth prompt and "Try it out" targeting `localhost`.
  - `cdk synth` / nag was previously run against the earlier `.yaml`-import version, not this JSON-import version — re-run synth to confirm the bundle still builds.
  - The full pre-commit suite did not run in my environment (local pnpm/Node version mismatch: pnpm 11 requires Node ≥22.13, only Node 20.x installed).
- E2E (Playwright) not run — it targets a deployed `quality` environment.

## Impact
- **Blast radius:** backend (`shared-lib` OpenAPI generator + Swagger helper, one Lambda handler) and the local dev server. No front-end, schema, or infra-resource changes.
- **New dependencies:** none. **New network calls:** none new (Swagger UI CDN assets were already part of the feature).
- **Behaviour when deployed:** unchanged (spec content is identical; served as JSON instead of parsed-from-YAML). The fix restores local-dev parity.
- Adds a committed generated artifact `easy-genomics-api.json` (~6,984 lines / 315 KB) — hence the large insertion count; it is machine-generated, not hand-written. `generate-openapi.ts` now writes both yaml and json so they stay in sync.

## Additional Information
Route naming: the handler is `list-api-docs.lambda.ts` (→ `GET /easy-genomics/list-api-docs`) rather than `read-api-docs`, because the repo's convention treats `read-` as a single-item verb that auto-appends `/{id}`; `list-` is the parameter-less GET (as with `list-buckets`). Path is fully convention-derived — no custom route wiring.

## Checklist*
- [x] No new errors or warnings have been introduced. *(ESLint clean on changed files; full build not run locally — see Testing)*
- [x] All tests pass successfully and new tests added as necessary. *(swagger-ui + openapi-guard suites pass; added a no-mutation test)*
- [ ] Documentation has been updated accordingly. *(N/A — no user-facing prose docs; the OpenAPI spec artifacts were regenerated)*
- [x] Code has been commented in particularly hard-to-understand areas. *(why JSON over YAML, the clone rationale, and the local Host-header fallback)*
](https://dept-au.atlassian.net/browse/EGV-214)


kevin-loftus-dept (approved):
LGTM

PR #844, opened 2026-07-16
fix: gate org view and authorize list-organization-users-details

ignacionistal:
## Type of Change*
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description

- Gate <EGOrgView> with v-if so unauthorized users never mount the org page or fetch members before redirect.
- Enforce auth on list-organization-users-details (org admin / lab manager / system admin; userId queries filtered accordingly)

## Testing*
Tested manually on development.

## Checklist*

### DIFF
--- a/packages/back-end/src/app/controllers/easy-genomics/organization/user/list-organization-users-details.lambda.ts
+++ b/packages/back-end/src/app/controllers/easy-genomics/organization/user/list-organization-users-details.lambda.ts
@@ -1,13 +1,62 @@
 import { buildErrorResponse, buildResponse } from '@easy-genomics/shared-lib/lib/app/utils/common';
+import { InvalidRequestError, UnauthorizedAccessError } from '@easy-genomics/shared-lib/lib/app/utils/HttpError';
 import { OrganizationUser } from '@easy-genomics/shared-lib/src/app/types/easy-genomics/organization-user';
 import { OrganizationUserDetails } from '@easy-genomics/shared-lib/src/app/types/easy-genomics/organization-user-details';
-import { OrganizationAccess, User } from '@easy-genomics/shared-lib/src/app/types/easy-genomics/user';
-import { APIGatewayProxyResult, APIGatewayProxyWithCognitoAuthorizerEvent, Handler } from 'aws-lambda';
+import {
+  LaboratoryAccessDetails,
+  OrganizationAccess,
+  OrganizationAccessDetails,
+  User,
+} from '@easy-genomics/shared-lib/src/app/types/easy-genomics/user';
+import {
+  APIGatewayProxyEvent,
+  APIGatewayProxyResult,
+  APIGatewayProxyWithCognitoAuthorizerEvent,
+  Handler,
+} from 'aws-lambda';
 import { OrganizationUserService } from '@BE/services/easy-genomics/organization-user-service';
 import { UserService } from '@BE/services/easy-genomics/user-service';
+import { validateOrganizationAdminAccess, validateSystemAdminAccess } from '@BE/utils/auth-utils';
 
 const organizationUserService = new OrganizationUserService();
 const userService = new UserService();
+
+/**
+ * True if the caller is an active Lab Manager for any laboratory in the org.
+ * Lab managers need org user listings when adding users to a lab.
+ */
+function validateAnyLaboratoryManagerAccess(event: APIGatewayProxyEvent, organizationId: string): boolean {
+  try {
+    const orgAccessMetadata: string | undefined = event.requestContext.authorizer?.claims?.OrganizationAccess;
+    if (!orgAccessMetadata) {
+      return false;
+    }
+
+    const organizationAccess: OrganizationAccess = JSON.parse(orgAccessMetadata);
+    const organizationAccessDetails: OrganizationAccessDetails | undefined = organizationAccess[organizationId];
+    if (!organizationAccessDetails?.LaboratoryAccess) {
+      return false;
+    }
+
+    return Object.values(organizationAccessDetails.LaboratoryAccess).some(
+      (lab: LaboratoryAccessDetails) => lab.Status === 'Active' && lab.LabManager === true,
+    );
+  } catch (error) {
+    console.error(error);
+    return false;
+  }
+}
+
+function canListUsersByOrganizationId(
+  event: APIGatewayProxyWithCognitoAuthorizerEvent,
+  organizationId: string,
+): boolean {
+  return (
+    !!validateSystemAdminAccess(event) ||
+    !!validateOrganizationAdminAccess(event, organizationId) ||
+    validateAnyLaboratoryManagerAccess(event, organizationId)
+  );
+}
 
 export const handler: Handler = async (
   event: APIGatewayProxyWithCognitoAuthorizerEvent,
@@ -18,7 +67,7 @@ export const handler: Handler = async (
     const organizationId: string | undefined = event.queryStringParameters?.organizationId;
     const userId: string | undefined = event.queryStringParameters?.userId;
 
-    const organizationUsers: OrganizationUser[] = await listOrganizationUsers(organizationId, userId);
+    const organizationUsers: OrganizationUser[] = await listOrganizationUsers(event, organizationId, userId);
 
     if (organizationUsers.length === 0) {
       return buildResponse(200, JSON.stringify([]), event);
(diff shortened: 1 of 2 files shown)


diegomayorga-dept:
# PR Review: fix: gate org view and authorize list-organization-users-details (#844)

## Summary
- 1 blocker, 7 major, 1 minor
- Top concerns: cross-tenant `OrganizationAccess` data leak on userId lookup; zero test coverage for new authorization branches; PR title/body missing ticket reference and breaking-change disclosure

## Correctness (1)

### major — packages/back-end/src/app/controllers/easy-genomics/organization/user/list-organization-users-details.lambda.ts:84-87

The new userId-only authorization branch correctly filters *which membership rows* are returned (only orgs the caller administers), but the response-enrichment code that builds each row's `OrganizationAccess` field only scopes to a single org when the **request's** `organizationId` query param is present:

```ts
const organizationAccess: OrganizationAccess | undefined =
  user.OrganizationAccess && organizationId
    ? { [organizationId]: user.OrganizationAccess[organizationId] }
    : user.OrganizationAccess;
```

For a userId-only request that param is always undefined, so `organizationAccess` falls through to the target user's FULL cross-org access map, including organizations/lab roles the caller has no authorization to see. Example: an org admin of Org A looks up a user who is also a member of Org B; the authorization filter correctly returns only the Org A membership row, but that row's embedded `OrganizationAccess` payload still contains Org B's `OrganizationAdmin`/`LaboratoryAccess` details, leaking cross-tenant data in the API response even though the UI only renders the Org A view. This directly contradicts the PR's stated goal of "userId queries filtered accordingly."

**Suggestion:** When building `organizationAccess` for each row, always scope to `orgUser.OrganizationId` (the specific membership being emitted) rather than only when the request's `organizationId` query param happens to be set, e.g. `user.OrganizationAccess ? { [orgUser.OrganizationId]: user.OrganizationAccess[orgUser.OrganizationId] } : undefined`.

## Conventions (1)

### major — packages/back-end/src/app/controllers/easy-genomics/organization/user/list-organization-users-details.lambda.ts:28

New authorization logic (`validateAnyLaboratoryManagerAccess`, `canListUsersByOrganizationId`) is defined directly in the `.lambda.ts` controller instead of `packages/back-end/src/app/utils/auth-utils.ts`, where every other `validate*Access` helper (`validateSystemAdminAccess`, `validateOrganizationAdminAccess`, `validateLaboratoryManagerAccess`, `validateOrganizationAccess`) lives. `validateAnyLaboratoryManagerAccess` also re-implements the same `OrganizationAccess`-claims JSON-parsing pattern already present multiple times in `auth-utils.ts`, duplicating rather than extending existing logic.

**Suggestion:** Move `validateAnyLaboratoryManagerAccess` into `auth-utils.ts` alongside the other `validate*Access` functions (consistent with the established pattern), and keep `canListUsersByOrganizationId` as a small handler-local composition of the imported validators, or also relocate it if similar composite checks exist elsewhere.

## Tests (2)

### blocker — packages/back-end/src/app/controllers/easy-genomics/organization/user/list-organization-users-details.lambda.ts:114

`listOrganizationUsers` now branches on five distinct authorization outcomes (system admin, org admin, lab manager, self-lookup, and the org-admin-filtered multi-org userId case) plus an unauthorized-throw path, but the file has zero test coverage — no test file exists for `list-organization-users-details.lambda.ts` at all. This is a direct security-relevant precedent gap: the near-identical sibling handler `list-laboratory-users-details.lambda.ts` has a full test file that imports and directly unit-tests its equivalent helper (`listLaboratoryUsers`) across every query-parameter combination plus invalid/failure branches, and `add-laboratory-user.lambda.test.ts` establishes the pattern for testing `validateOrganizationAdminAccess`-gated authorization (mocking the auth-utils functions and asserting both allow and deny outcomes). None of that precedent was followed here for logic that gates access to org member PII.

**Suggestion:** Add `packages/back-end/test/app/controllers/easy-genomics/organization/user/list-organization-users-details.lambda.test.ts` mirroring `list-laboratory-users-details.lambda.test.ts`: export `listOrganizationUsers` (and ideally `canListUsersByOrganizationId`/`validateAnyLaboratoryManagerAccess`) for direct testing, mock `validateSystemAdminAccess`/`validateOrganizationAdminAccess` the way `add-laboratory-user.lambda.test.ts` does, and cover: system admin listing by organizationId, org admin listing by organizationId, lab manager listing by organizationId, non-manager/non-admin denied (`UnauthorizedAccessError`), system-admin/self userId lookup, org-admin filtered userId lookup with 0 and >0 authorized results, and both organizationId+userId supplied (`InvalidRequestError`).

### major — packages/back-end/src/app/controllers/easy-genomics/organization/user/list-organization-users-details.lambda.ts:28

`validateAnyLaboratoryManagerAccess` wraps `JSON.parse(orgAccessMetadata)` in a try/catch that swallows parse errors and returns false, and separately falls back to false for missing `OrganizationAccess` claim, missing `LaboratoryAccess`, and empty/all-inactive `LaboratoryAccess` maps — none of these defensive branches have a test verifying they actually degrade to "deny" rather than throwing an unhandled error up through `canListUsersByOrganizationId`.

**Suggestion:** Add unit tests for `validateAnyLaboratoryManagerAccess` covering: malformed/non-JSON `OrganizationAccess` claim string, missing `OrganizationAccess` claim, organization entry with no `LaboratoryAccess`, `LaboratoryAccess` present but all `Status !== 'Active'` or `LabManager !== true`, and the true-returning case where one lab has `Status: 'Active'` and `LabManager: true`.

## Simplification (1)

### minor — packages/back-end/src/app/controllers/easy-genomics/organization/user/list-organization-users-details.lambda.ts:28

`validateAnyLaboratoryManagerAccess` duplicates the `OrganizationAccess`-claims parsing and try/catch/`console.error` pattern already implemented in `auth-utils.ts`'s `validateLaboratoryManagerAccess` (checks a specific laboratoryId) and `getLaboratoryAccessLaboratoryIds` (returns active lab IDs for an org). This is effectively a 5th near-identical reimplementation of that parsing logic scattered across `auth-utils.ts`, now duplicated locally in the lambda instead of being centralized.

**Suggestion:** Reuse `getLaboratoryAccessLaboratoryIds(event, organizationId)` from `auth-utils.ts` to get the active lab IDs, then check `LabManager` on each via the org's `LaboratoryAccess` map (or extend `validateLaboratoryManagerAccess` to accept an optional laboratoryId and iterate over `getLaboratoryAccessLaboratoryIds` when omitted), keeping the claims-parsing logic in one place.

## PR Hygiene (4)

### major — PR:0

Title "fix: gate org view and authorize list-organization-users-details" uses a conventional-commit-style "fix:" prefix and has no ticket reference (no EGV-xxx or similar), diverging from the repo's ticket-prefix title convention.

**Suggestion:** Rename to lead with the ticket key, e.g. "EGV-XXX Gate org view and authorize list-organization-users-details".

### major — PR:0

Body has no link to a ticket/issue and doesn't explain why this authorization gap existed or how it was discovered (e.g. security report, audit finding, incident) — only describes what was changed.

**Suggestion:** Add a ticket link and a brief root-cause note (how the missing auth check was found) to the Description section.

### major — PR:0

Test plan is just "Tested manually on development" with no enumerated scenarios or roles. This is a security-sensitive authorization fix (org admin / lab manager / system admin gating), so the specific roles and negative-path (unauthorized access denied) tests exercised should be listed.

**Suggestion:** Enumerate the roles tested (org admin, lab manager, system admin, and an unauthorized user) and confirm both allowed and denied cases were verified.

### major — PR:0

This PR tightens `list-organization-users-details` from open-to-any-authenticated-user to role-gated, which could break existing callers relying on the previous open access. The body does not mention this as a potential breaking change or note any rollout/coordination consideration.

**Suggestion:** Add a note under Description or a "Breaking Changes" callout stating that previously-permitted callers without the required role will now receive an authorization error, and confirm no other consumers depend on the old behavior.

---

Security and Breaking-changes dimensions returned zero findings after independent deep-repo verification (Cognito authorizer chain, claim provenance, and all four query-param branches were traced and confirmed clean) — omitted above per the report template.

☑️ (issues found)

kevin-loftus-dept (approved):
### Non-blocking follow-up: scope OrganizationAccess on userId queries

**Finding**

Auth correctly filters membership rows for org admins on `?userId=…`, but the response still attaches the target user’s **full** `OrganizationAccess` (all orgs/labs) when `organizationId` is absent.

In `list-organization-users-details.lambda.ts`, the mapping is effectively:

- if `organizationId` is set → `{ [organizationId]: user.OrganizationAccess[organizationId] }`
- else (`userId` path) → `user.OrganizationAccess` (full map)

So an org admin of Org A who can list a user shared with A may still see that user’s lab roles / membership in Org B via the API payload (e.g. `EGUserAccess` → `usersDetailsByUserId`).

This is **pre-existing** response shaping (same on `development`). This PR still improves security by restricting who can call the endpoint; treating as follow-up, not a merge blocker.

**Proposed fix**

On the `userId` path, scope `OrganizationAccess` to orgs the caller is allowed to see (mirror the membership filter):

- system admin / self → keep full map (current behavior)
- org admin → only include keys where `validateOrganizationAdminAccess(event, orgId)` is true  
  (or, per row, only `{ [orgUser.OrganizationId]: user.OrganizationAccess[orgUser.OrganizationId] }`)

That matches the existing `organizationId` path, which already returns a single-org map.

PR #854, opened 2026-07-27
feat: (egv-225) lab 02 email notifications

diegomayorga-dept:
## Title

[LAB-02] Email notifications for pipeline run completion (v1)

## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description

Adds opt-in email notifications when a pipeline run reaches a terminal state (COMPLETED/FAILED/etc.), so lab techs don't have to keep the lab dashboard open to know a run finished. Follows the spike in `notification-investigation.md` (GO decision: SNS→SQS→SES architecture, per-user opt-in default off, per-event over daily digest).

**Reliability prerequisite — active-run poller:** run status today only refreshes while a user has the dashboard open (`EGLabView.vue`'s 2-minute front-end poll). New `process-poll-active-runs.lambda.ts` runs on an EventBridge schedule (every 2 min), finds every non-terminal run via a new sparse `PollStatus` GSI (`easy-genomics-api-stack.ts`, `LaboratoryRunService.queryActiveForPolling`), and re-enqueues a status check on the existing `laboratory-run-update-topic` — so a terminal transition is detected even with no browser open.

**Trigger + idempotency:** `process-update-laboratory-run.lambda.ts` (existing) now calls `LaboratoryRunService.markTerminalNotified` on a genuine non-terminal → terminal transition — a conditional DynamoDB write (`ConditionExpression: attribute_not_exists(NotifiedAt)`) that guarantees exactly-once publish even under duplicate/retried status-check messages. Only the write that wins the condition publishes to a new `laboratory-run-notification-topic` (SNS FIFO → new SQS queue + DLQ, `easy-genomics-nested-stack.ts`).

**Delivery:** new `process-notify-laboratory-run-completion.lambda.ts` consumes the queue and delegates to `NotificationService.notifyRunCompletion`, which resolves recipients (run owner if `NotifyOnOwnRuns`, plus lab members who opted into `NotifyOnLabRuns`, deduped by `UserId`), respects each recipient's `NotificationEventFilter` (all terminal states vs. failures only) and the lab-level `NotificationsEnabled` kill switch, and sends one templated email per recipient via a new `SesService.sendRunCompletionEmail` / `RunCompletionEmailTemplate`. A single recipient's send failure is caught and logged, never fatal to the batch.

**Preferences:** new fields `Laboratory.NotificationsEnabled`, `LaboratoryUser.NotifyOnLabRuns`, `User.NotifyOnOwnRuns`/`NotificationEventFilter` (shared-lib schemas + their hand-maintained `.d.ts` type mirrors, which this codebase's OpenAPI generator reads separately from the Zod schemas). New self-service endpoint `update-laboratory-user-notification-preference.lambda.ts` lets a user toggle their own per-lab opt-in — deliberately a new, narrowly-scoped file rather than widening the admin-gated `edit-laboratory-user.lambda.ts`, since `UserId` is derived only from the Cognito claim, never the request body. Front-end gained two account-level toggles on `profile.vue` (email me on my own runs; all terminal states vs. failures only), fetched via a direct `getUser()` call into local page state rather than the global user store (that store mirrors a JWT-claim mechanism this change doesn't touch).

**Migration:** one-time `backfill-laboratory-run-attributes.ts` script (run once before deploy) does two passes: stamps `PollStatus='ACTIVE'` on runs already in flight at deploy time, and stamps a presence-only `NotifiedAt` sentinel on all pre-existing terminal runs — the latter specifically to prevent a stale "run finished" email firing the first time someone reopens an old run after this feature ships (the same healing branch that recovers a stranded notification is also reachable via the existing duration-backfill status-check path for legacy runs).

**Out of scope / follow-up:** SMS, Slack, or in-app notifications; notifications for in-progress/non-terminal events; daily digest delivery; run cost in the notification payload; front-end UI for the per-lab opt-in and the lab-manager kill switch (both API-complete from this PR, no UI yet — deliberately deferred).

## Testing

Full back-end suite: 73 suites / 541 tests passing, including new unit tests for `LaboratoryRunService.markTerminalNotified`/`queryActiveForPolling`, `NotificationService` (6 scenarios: kill switch, owner opt-in/out, lab-member opt-in + owner dedup, per-recipient failure isolation, `failures_only` filter), the poller and sender Lambdas, the self-service preference endpoint (including the "UserId never from the request body" security case), and CDK infra wiring assertions (SNS/SQS/DLQ, IAM grants, EventBridge schedule).

Not automated: actual SES email delivery. `TESTING.md` has a manual verification entry (SES simulator addresses) logged as "not yet executed" — no deployed environment was available during implementation to run it. **To verify:** deploy to `quality`, set a test user's `NotifyOnOwnRuns=true`, trigger a run to completion, and confirm delivery via `success@simulator.amazonses.com`; separately confirm a `bounce@simulator.amazonses.com` recipient doesn't block other recipients in the same batch.

The one-time backfill script (`backfill-laboratory-run-attributes.ts`) has no automated test, matching every sibling script in `packages/back-end/scripts/` — verified locally that it loads env, parses flags, and reaches the DynamoDB call before failing on missing credentials (expected in this sandboxed environment); **to verify:** run with `--dry-run` against `quality` before the real run.

Front-end toggles were not manually verified in a browser — no deployed backend/env config was available in this environment. **To verify:** load `/profile`, confirm the toggle and select reflect the current user's stored preferences, flip them, and confirm a `PUT /user/update-user-request/{id}` call fires with the expected body.

## Impact

- **New AWS resources:** `PollStatus_Index` GSI on `laboratory-run-table`; SNS FIFO topic + SQS FIFO queue + DLQ (`laboratory-run-notification-*`); 3 new Lambdas (2 SQS/schedule-triggered, 1 HTTP); 1 new EventBridge rule (2-minute poll); 1 new SES template.
- **CloudFormation resource budget:** the shared nested stack grew from 421 → 436 of the 500 hard limit across this feature (CI-enforced warn=400/fail=450 — still under fail threshold, but worth tracking on future PRs to this stack).
- **Schema changes:** additive only (new optional fields on `Laboratory`, `LaboratoryUser`, `User`, `LaboratoryRun`); no existing field changed or removed; no data migration required for existing rows other than the one-time backfill script above.
- **New dependencies:** none.
- **Behavior change for existing endpoints:** `process-update-laboratory-run.lambda.ts`'s backfill-healing branch (which already heals `TerminalAt`/`ExpiresAt`/`RunDurationSeconds` on already-terminal runs) now also heals a missing `NotifiedAt` — this is what the migration script's second pass exists to make safe for pre-existing data.
- **Front-end:** additive UI on `profile.vue` only; no existing page behavior changed.

## Additional Information

Recurring gotcha hit three times during implementation and worth flagging for reviewers: this repo has hand-maintained `.d.ts` type mirrors under `packages/shared-lib/src/app/types/easy-genomics/*.d.ts`, separate from the Zod schema files, that the OpenAPI generator reads instead of `z.infer<...>`. Every new schema field in this PR required updating both; all four affected `.d.ts` files are included in this diff.

Design doc: `superpowers/specs/2026-07-24-lab-02-notifications-implementation-design.md`. Implementation plan (12 tasks, executed via subagent-driven development with a task-level review after each and a 3-round whole-branch review at the end — the whole-branch pass caught a missing `sns:Publish` IAM grant on the poller that would have made the feature silently inert in production, plus the stale-notification migration gap described above): `superpowers/plans/2026-07-24-lab-02-email-notifications.md`.

## Checklist

- [ ] Documentation has been updated accordingly. *(TESTING.md updated with the manual SES-verification entry; no other user-facing docs affected)*
- [ ] Code has been commented in particularly hard-to-understand areas. *(comments added only where non-obvious — e.g. idempotency guard rationale, sparse-GSI rationale — not a blanket pass)*


kevin-loftus-dept:
## Review notes — two correctness issues

### 1. Lab Settings can silently disable notifications on save (blocking)

**Where:** `EGFormLabDetails.vue` (`defaultState` + `getLabDetails`) → `update-laboratory.lambda.ts`

**What's wrong:**
- Form `defaultState` initializes `NotificationsEnabled: false`.
- `getLabDetails` does `state.value = { ...state.value, ...withRetentionDefault }`.
- For labs where the attribute is absent (all existing labs today; create also never persists this field), the API payload has no `NotificationsEnabled`, so the merge **keeps** the leftover `false` on `state`.
- `uneditedLabDetails` is built only from the API payload → `NotificationsEnabled` is `undefined`.
- Dirty check is `false !== undefined` → form looks dirty immediately.
- Next Save writes `NotificationsEnabled: false`.

**Backend semantics:** `NotificationService` only skips when `NotificationsEnabled === false`; missing/`undefined` means enabled (kill-switch default on). So this Save flips labs from “enabled by default” to “hard off” without the user intending to.

**Repro sketch:** Open Lab Settings on a lab that has never set this field → Save any change (or Save while the false dirty state is present) → notifications for that lab stop.

**Proposed fix:**
1. Treat absent as enabled in the form: after load, normalize  
   `NotificationsEnabled: labDetails.NotificationsEnabled !== false`  
   (or default `true`) on **both** `state` and `uneditedLabDetails`.
2. Change `defaultState` to `true` / `undefined` instead of `false`, matching kill-switch semantics.
3. Add a dirty-check norm for this field (like `AwsHealthOmicsNetworkingMode`) so `undefined` and `true` compare equal if you keep optional storage.
4. Optional hardening: have `create-laboratory` persist an explicit default (`true`) so FE/BE stay aligned.

---

### 2. Mark-then-best-effort-publish can permanently drop a notification (high)

**Where:** `process-update-laboratory-run.lambda.ts` — `markTerminalNotified` then `safePublishForNotification`

**What's wrong:**
1. Winner of `attribute_not_exists(NotifiedAt)` stamps `NotifiedAt` and clears `PollStatus`.
2. `safePublishForNotification` swallows SQS errors (`console.warn` + continue).
3. Healing / retries only re-publish when `NotifiedAt == null`.

If the conditional write succeeds and the subsequent `SendMessage` fails (or env URL is empty), that run **never** emails and the poller won't pick it up again (`PollStatus` already removed). This is worse than `safePublishForClassification`, which can still retry because nothing is marked done before enqueue.

**Proposed fix (pick one):**

**A (preferred, small):** Keep mark-first for exactly-once, but on publish failure **compensate**: clear `NotifiedAt` and restore `PollStatus: 'ACTIVE'` (or rethrow so the status-check message retries *and* compensate before returning). Then the existing healing/poller path can retry.

**B:** Enqueue first, then `markTerminalNotified`. Use a stable FIFO `MessageDeduplicationId` (e.g. `${RunId}-terminal-notify`) to limit duplicates. Accept rare duplicate email if mark fails after a successful enqueue.

**C:** Two-phase attribute (`NotificationPending` → `NotifiedAt`) set only after the sender succeeds — more moving parts.

Also worth asserting in tests that a rejected `sendMessage` after a successful mark leaves the run retryable (NotifiedAt absent / PollStatus ACTIVE).

kevin-loftus-dept (approved):
See my PR comment

diegomayorga-dept:
> See my PR comment

Yes — both issues [review comment](https://github.com/dept/easy-genomics/pull/854#issuecomment-5171383045) were addressed, in commit 52d27d20 ("fix: address code review on PR #854 notification bugs"),

PR #879, opened 2026-08-11
fix: persist run status before cost capture so large Omics runs don't hang at RUNNING

kevin-loftus-dept:
## Title*
Fix large HealthOmics runs stuck at RUNNING when status-check cost capture times out

## Type of Change*
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Omics runs that finish on the platform could remain `RUNNING` in Easy Genomics because `process-update-laboratory-run` awaited HealthOmics cost capture (`listAllRunTasks`) **before** writing the terminal status to DynamoDB.

For large runs (observed: 5,000+ tasks, ~2 minutes to page `ListRunTasks`), that work exceeds the Lambda’s 30s timeout. The invocation fails, SQS redrives the same message, and DynamoDB never gets `COMPLETED`/`FAILED`/etc.—so the UI keeps polling and shows the run as in progress.

This change:
- Persists status / terminal metadata / notification side effects **before** cost capture (status-change and backfill paths)
- Time-boxes cost capture to 8s (`Promise.race`) so it cannot block the pipeline
- Sets `callbackWaitsForEmptyEventLoop = false` so abandoned in-flight `ListRunTasks` requests do not hold the invoke open until timeout and redrive SQS
- Updates unit tests for the status-then-cost write order
- Bumps the flaky `backfill-laboratory-run-attributes` suite timeout (resetModules under full parallel runs)
- Allows `.env.nuxt.local` to override keys already loaded by `nuxt-load-settings` (`dotenv` `override: true`) for local FE config

Related: EGV-244

## Testing*
- Updated `process-update-laboratory-run.lambda.test.ts` for status-first then cost follow-up update; cost-failure path still swallows errors and writes status
- Re-ran `process-update-laboratory-run` and `backfill-laboratory-run-attributes` Jest suites locally
- Runtime verification against quality: local invoke of the fixed handler wrote `Status=COMPLETED` for a stuck Omics run; subsequent status-check reported `runsToProcessCount: 0`

## Impact
- Terminal status updates no longer depend on finishing a full Omics task census
- Large runs may temporarily lack task-based `RunCostOutcome` if capture exceeds 8s (manual `backfill-run-cost-outcomes` or later capture attempts can fill it; daily `process-sync-run-costs` is Cost Explorer `BilledCost`, not this path)
- Progress reporting while a run is still non-terminal is unchanged (separate `listAllRunTasks` path; not under the 8s budget)
- Requires deploying `process-update-laboratory-run` for the fix to apply in AWS; local SQS env alone only affects enqueueing

## Additional Information
- Branch also includes a small front-end local-env override fix and a test-timeout bump unrelated to the cost/status race—call out in review if you’d rather split those
- Quality had built up a large backlog on `laboratory-run-update-queue.fifo` because timed-out invocations kept redriving the same large-run messages. After this consumer is deployed, those messages should be able to persist terminal status and complete successfully so SQS can delete them instead of retrying forever

## Checklist*

### DIFF
--- a/packages/back-end/src/app/controllers/easy-genomics/laboratory/run/process-update-laboratory-run.lambda.ts
+++ b/packages/back-end/src/app/controllers/easy-genomics/laboratory/run/process-update-laboratory-run.lambda.ts
@@ -42,11 +42,30 @@ const ssmService = new SsmService();
 
 /**
  * Best-effort platform cost capture. Must never break the status-check pipeline.
+ *
+ * HealthOmics cost uses paginated ListRunTasks; large runs can take well over the
+ * Lambda timeout (observed: 5k+ tasks / ~2min). Cap wait time so status/notify
+ * writes always complete; uncaptured cost can be healed by process-sync-run-costs.
  */
+const COST_CAPTURE_BUDGET_MS = 8_000;
+
 async function safeCaptureRunCost(run: LaboratoryRun): Promise<LaboratoryRun['RunCostOutcome'] | undefined> {
   if (run.RunCostOutcome?.CostCapturedAt) return run.RunCostOutcome;
   try {
-    return await captureRunCostOutcome(run);
+    let timeoutId: ReturnType<typeof setTimeout> | undefined;
+    const result = await Promise.race([
+      captureRunCostOutcome(run),
+      new Promise<undefined>((resolve) => {
+        timeoutId = setTimeout(() => {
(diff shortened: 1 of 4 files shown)


ignacionistal (approved):
LGTM

PR #881, opened 2026-08-14
feat: deploy-gated migration runner

diegomayorga-dept:
## Summary

Adds a deploy-gated migration runner so routine, idempotent, low-blast-radius data migrations run automatically during back-end deploy (local and CI), at most once per environment, gated by explicit pre/post-CDK phases.

Today only `migrate-laboratory-s3-access-seed` runs automatically (post-CDK). Other one-time scripts (e.g. `backfill-laboratory-run-attributes`) are manual — a missed or mistimed run causes real bugs (in-flight runs invisible to the poller, spurious completion emails). This closes that gap for opt-in, safe-to-automate migrations while keeping expensive/Tier-3 scripts manual.

Spec: `docs/superpowers/specs/2026-08-13-deploy-gated-migration-runner-design.md` equivalent (kept outside this repo per project convention).

## What's included

- **Runner CLI** (`packages/back-end/scripts/run-deploy-migrations.ts`) — `--phase pre|post` (required), `--dry-run`, `--force <id>`. Filters an opt-in registry by phase, skips already-applied ids, fails fast (and leaves the ledger untouched) on the first migration failure.
- **Registry** (`scripts/deploy-migrations/registry.ts`) — statically imports each migration's `main()`, tagged with a stable id and phase. Currently two entries: `backfill-laboratory-run-attributes` (phase `pre`) and `migrate-laboratory-s3-access-seed` (phase `post`).
- **Ledger** (`scripts/deploy-migrations/ledger.ts`) — SSM Parameter Store at `/${NAME_PREFIX}/deploy-migrations/applied`, a JSON map of `{ [id]: { appliedAt } }`. Missing parameter = empty ledger. Best-effort read-modify-write-retry loop (SSM has no compare-and-swap).
- **Script adaptations** — both registered scripts now throw instead of calling `process.exit()`, so the runner controls ledger writes and exit codes.
- **Deploy wiring** (`.projenrc.ts`) — `deploy` now runs: `cdk bootstrap → preflight-deletion-protection → run-deploy-migrations --phase=pre → cdk deploy --all → run-deploy-migrations --phase=post`. Both local `pnpm run deploy`/`build-and-deploy` and CI (`cicd-build-deploy-back-end`) pick this up with no workflow YAML changes.
- **Docs** — `scripts/README.md` (how to add a migration, phase rules, ledger path, flags) and a Tier 1/2 note in `docs/deployment/upgrading.md`.

## Fixes found during final review

A whole-branch review caught two issues invisible to per-task review, both fixed and covered by regression tests:

- **CI would fail every deploy's pre-phase**: `backfill-laboratory-run-attributes.ts` required a `REGION` env var that CI never sets (only `AWS_REGION`). Now reconciles either direction.
- **Greenfield deploys were permanently blocked**: the backfill's table scan had no guard for a not-yet-existent table, unlike the existing S3 seed script's pattern. Added the same `ResourceNotFoundException` guard.
- **`--force` without an id silently fell through to a full run** instead of erroring — now throws, and `parseArgs` has direct test coverage.

## Testing

- `pnpm test` in `packages/back-end`: 112 suites / 826 tests passing.
- New unit tests cover: ledger parse/merge/missing-param/retry-then-throw; runner phase filtering, ledger skip, fail-fast + no ledger write on failure, `--dry-run` no-op, `--force` re-apply and phase-scoping, `--force`-without-id error; both adapted scripts' throw-on-failure and greenfield-guard behavior.
- Manual QA plan (non-prod environment) is documented in the design spec — not yet executed against a live AWS account; recommended before this ships to an environment with real data.

## Known follow-ups (not blocking, called out for reviewer awareness)

- Verify the CI OIDC role and local deploy identities have `ssm:GetParameter`/`ssm:PutParameter` on `/${NAME_PREFIX}/deploy-migrations/applied` — couldn't confirm IAM policy from this repo alone.
- `migrate-laboratory-s3-access-seed.ts` resolves `NAME_PREFIX` eagerly at module-import time (pre-existing behavior, unchanged here) — confirmed safe for both CI and local deploy paths, but a future cleanup could move it inside `main()`.
- A few Minor polish items from the final review (ledger parse-error message clarity, a stray trailing CLI arg from `build-and-deploy`, a doc wording nit) were intentionally deferred as non-blocking.


ignacionistal (approved):
LGTM

kevin-loftus-dept (approved):
Nice work!

PR #923, opened 2026-08-26
fix: stop spurious Seqera error toast when cancelling a run

ignacionistal:
## Type of Change*
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
QA reported that cancelling a HealthOmics run shows the error toast
"Failed to load Seqera runs. Please refresh." even though no Seqera run is
involved.
The cancel itself succeeds. The error comes from the refresh that runs
immediately afterwards: `handleCancelDialogAction` unconditionally called both
`getSeqeraRuns()` and `getOmicsRuns()`, regardless of the cancelled run's
platform or whether the lab can reach either platform. On a lab that has
`NextFlowTowerEnabled` set but no Personal Access Token stored, the Seqera fetch
hits `GET /nf-tower/workflow/list-workflows`, which throws
`LaboratoryAccessTokenUnavailableError` (or `MissingNextFlowTowerAccessError`
when the flag is off). `runStore.loadSeqeraRunsForLab` catches that and toasts.
This is not a regression from #916. Those two refresh calls have been unguarded
since the lab view was extracted into a component (904265a8). Previously the
cancel request itself failed too, so the Seqera toast appeared alongside
"Failed to cancel run" and blended in. Now that cancel works, it's the only
error left on screen, which made it look new.
The fix gates each refresh on the lab's capability flags, reusing the same
conditions this component already uses to decide which tabs to show and which
data to fetch in the `watch(lab)` handler. The Omics refresh is gated
symmetrically because the mirror-image bug exists: cancelling a Seqera run in a
Seqera-only lab would toast "Failed to load Omics runs. Please refresh."

## Testing*
1. On a lab with HealthOmics enabled and Seqera enabled but **no** PAT, start a
   run, cancel it, and confirm no Seqera error toast appears.
2. On a lab with both platforms fully configured, cancel a run on each platform
   and confirm both still cancel and refresh correctly.

## Impact
Front-end only, scoped to the post-cancel refresh path. Removes one HTTP request
on labs that can't serve it. No API, schema, or infrastructure changes.

## Additional Information
[Any additional information that reviewers should be aware of.]

## Checklist*

### DIFF
--- a/packages/front-end/src/app/components/EGLabView.vue
+++ b/packages/front-end/src/app/components/EGLabView.vue
@@ -264,6 +264,15 @@
 
   type LaboratoryRunTableItem = LaboratoryRun & { lastUpdated: string; searchIndex: string };
 
+  const TERMINAL_STATUSES = ['FAILED', 'SUCCEEDED', 'CANCELLED', 'COMPLETED', 'DELETED'];
+
+  /**
+   * Runs cancelled during this session. Cancelling only signals the compute platform; the
+   * LaboratoryRun record is updated later by the status check pipeline, so these are shown as
+   * cancelled until the server reports a terminal status of its own.
+   */
+  const locallyCancelledRunIds = ref<Set<string>>(new Set());
+
   const runsTableColumns = [
     { key: 'RunName', label: 'Run Name', sortable: true },
     { key: 'WorkflowName', label: 'Workflow name', sortable: true },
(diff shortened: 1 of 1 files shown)


diegomayorga-dept (approved):
LGTM

PR #933, opened 2026-09-14
fix: drop favourite workflows that no longer exist

ignacionistal:
## Title*
Fix deleted workflows remaining in Favorite Workflows

## Type of Change*
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Reported by client: *"If a workflow is deleted, it remains in the user's Favorites list
with no way to unfavorite or remove it."*

`FavouriteWorkflows` is a snapshot stored on the user record (`WorkflowId`,
`WorkflowName`, `Description`, `Platform`, `LaboratoryId`). `EGDashboard` filtered that
array by `LaboratoryId` alone and never checked it against the laboratory's actual
workflows, so an entry outlived the workflow it pointed at. The only unfavourite control
was the star on the HealthOmics Workflows tab, and that row no longer exists once the
workflow is deleted — hence "cannot be removed".

Changes, all front-end and **no DynamoDB migration**:

- **`utils/favourite-workflows.ts` (new)** — pure logic that classifies each favourite
  against what the session knows per platform: `available`, `missing`, `unknown` (the
  list fetch failed or was incomplete) or `platform-disabled`.
  `selectVisibleFavouriteWorkflows` drives display and hides `missing` /
  `platform-disabled` entries. Favourites are **not** deleted from the user record: a
  live list cannot distinguish deletion from access revocation or a failed shared-workflow
  fetch, so auto-pruning would be irreversible and incorrect.
- **`stores/favouriteWorkflows.ts` (new)** — single source of truth for favourites, with
  `load` (coalesced; returns whether a fresh server snapshot was applied),
  `toggleFavourite`, and `save`. Deliberately not persisted. `save` is gated behind a
  `loaded` flag so a failed user fetch cannot write a one-element array as the user's
  complete cross-lab list. `toggleFavourite` re-fetches immediately before writing and
  **aborts** if that re-read fails, so another tab's additions are not overwritten by a
  stale snapshot. A `loadGeneration` counter discards in-flight loads after logout/reset.
- **`stores/omicsWorkflows.ts`** — tracks `hasCompleteWorkflowList` per lab. A swallowed
  `listShared` failure still shows private workflows, but the list is marked incomplete so
  the dashboard does not treat it as `loaded` (which would hide every shared favourite).
- **`EGDashboard.vue`** — validates favourites against the live pipeline/workflow lists,
  hides ones that are no longer present, and adds a star button so any still-visible
  favourite can be removed from where it is displayed. Omics is only `loaded` when both
  private and shared lists resolved.
- **`EGLabView.vue`** — reads and toggles through the store instead of its own copy.

Stale ids may remain in DynamoDB until the user explicitly unfavourites a still-visible
workflow (or until a future deliberate cleanup). They no longer appear on the dashboard
once the underlying workflow is gone from the lab's live list.

## Testing*
- `pnpm test` in `packages/front-end`: suite green; **21** new tests across:
  - `test/app/utils/favourite-workflows.test.ts` — available vs. missing vs. unverifiable
    lists, per-platform matching, other labs untouched, disabled platform hidden.
  - `test/app/stores/favouriteWorkflows.test.ts` — load coalescing and boolean result,
    load-failure retry, refresh failure keeps prior state without toast, toggle aborts
    when re-read fails (no stale write), mid-flight `reset()` discard, no-signed-in-user
    guard, toggle add/remove with fresh server snapshot.
  - `test/app/stores/omicsWorkflows.test.ts` — complete private+shared list vs. incomplete
    after shared failure.
- Changed TypeScript files clean under `eslint` and `prettier`.
- Manual check on a HealthOmics lab:
  1. Star a workflow → it appears under Favorite Workflows on the dashboard.
  2. Delete the workflow in HealthOmics → reload the dashboard → the favourite is
     **hidden** (no longer offered for launch).
  3. Reload again → it stays **hidden**. It may still exist on the user record; that is
     expected (no auto-prune). Confirm it does **not** reappear in the UI.
  4. Star-remove a favourite for a workflow that still exists → it is removed from the
     dashboard and from the user record.

## Impact
- Deleted or access-revoked workflows stop being offered for launch, so users no longer
  hit a failing run from a dead favourite.
- One fewer `list-user-self` request per lab load: `EGLabView` and `EGDashboard`
  previously fetched the user separately.
- Explicit star toggles still write the whole array through
  `PUT /user/update-user-request/{id}`; there is **no** unattended write on dashboard load.
- No back-end, schema, infrastructure or dependency changes.

## Checklist*


diegomayorga-dept:
## PR Review

Hides favourite workflows whose underlying workflow is no longer in the lab's live list, instead of deleting them from the user record.

The previous round's two blockers remain fixed: the prune is gone entirely, `omicsWorkflows.hasCompleteWorkflowList` prevents a swallowed `listShared` failure from reading as `'loaded'`, and `save()` is gated behind a `loaded` flag plus a `loadGeneration` counter. Test coverage has also filled in — `reset()` mid-flight, load-failure retry, and the no-signed-in-user guard are all now covered.

### Critical issues

- **PR description contradicts the implementation** (body vs. `packages/front-end/src/app/stores/favouriteWorkflows.ts`, `packages/front-end/src/app/components/EGDashboard.vue`). Unchanged since the first commit, it still describes `pruneMissingFavouriteWorkflows`, "prunes them from the user record after load", and "stale rows already in DynamoDB drain away by themselves" — all removed in `83d1f175`. The Testing section still tells QA to confirm the favourite "stays gone after a second reload (confirming the prune was persisted)", which now fails by design: the row stays in DynamoDB and is only hidden. QA will read that as a regression.

- **`toggleFavourite` writes a stale snapshot when its re-read fails** (`packages/front-end/src/app/stores/favouriteWorkflows.ts:91-105`). The comment says the `await this.load()` exists "so a favourite added in another tab is not overwritten by this session's older snapshot", but `load()` swallows its error and, once `loaded` is `true`, deliberately doesn't even toast — the test at `packages/front-end/test/app/stores/favouriteWorkflows.test.ts:78` asserts `toastError` is *not* called. Control then falls through to `save()`, which writes the pre-existing in-memory array as the user's complete list, silently dropping another tab's additions. The re-read test at line 139 only exercises the success path, so this consequence is untested. Have `load()` report failure to its caller and abort the toggle when the re-read didn't land.

☑️ (issues found)

diegomayorga-dept (approved):

PR #934, opened 2026-09-14
fix: hide estimated cost until a workflow has enough run history

ignacionistal:
## Title*
Hide the estimated cost metric until a workflow has enough run history

## Type of Change*
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [x] UI/UX improvement

## Description
Client-reported: the "Pending" state for estimated costs reads as a hung or
stalled job. Expected behaviour is to hide the estimated cost metric entirely
until at least 3 successful runs exist for that workflow.

The 3-run minimum was already enforced on the back end — `estimateComputeCostBand`
returns `estimateAvailable: false` below `MIN_COMPARABLE_RUNS_FOR_COST_ESTIMATE`
comparable successful runs, so no estimate is ever produced for a fresh workflow.
The bug was purely presentational: `EGRunCostRow` always rendered its row, so a
run with no history displayed the label "Estimated cost" beside placeholder copy
("Billed cost pending", "Cost estimate unavailable").

The row is now hidden unless there is a real amount to show: a billed total, a
captured platform compute outcome, or a pre-run estimate backed by the minimum
sample size. The rule lives in a new `run-cost-row-visibility` util (mirroring the
existing `run-progress-card-visibility` pattern) so it is unit-testable without
mounting the component, and the threshold is now a single shared constant used by
both the estimator and the UI instead of two hardcoded `3`s.

Two deliberate choices worth a reviewer's attention:
- The row still shows "Calculating estimate…" while the request is in flight.
  Hiding it during loading would make the row pop in with no warning, and a brief
  accurate loading message is not the state the client complained about.
- The existing cost-source cascade and its "pending" copy are left in place. If
  the visibility rule ever lets an edge case through, the row degrades to the
  previously understood text rather than a bare dash.

## Testing*
- 18 new unit tests in `test/app/utils/run-cost-row-visibility.test.ts` covering
  the boundary at 3 comparable runs, zero-cost outcomes (a real $0.00 that must
  still display), storage-only outcomes, a stale pre-launch estimate on a
  launched run, and missing/failed estimate responses.
- Full front-end suite: 172 tests passing.
- `shared-lib` estimator suite (9 tests) and back-end `run-cost` suites (26
  tests) pass unchanged, confirming the constant extraction is behaviour-neutral.
- Prettier and ESLint clean on the changed files.

## Impact
Affects the three surfaces sharing `EGRunCostRow`: Review & Launch for AWS
HealthOmics and Seqera Cloud, and the run detail page. No API, schema, or
persisted-data changes, and no migration.

Behavioural note for the reviewer to confirm with the client: a completed run on
a brand-new workflow now shows no cost line at all until Cost Explorer syncs a
billed total or the platform compute outcome is captured. That follows directly
from "hide entirely", but it does mean silence where there used to be an
explanation.

## Additional Information
`MIN_COMPARABLE_RUNS_FOR_COST_ESTIMATE` intentionally lives in its own
dependency-free `run-cost-thresholds` module rather than in
`run-cost-estimation.ts`: the estimator imports node's `crypto`, which must not
be pulled into the client bundle (`ssr: false`, no node polyfills configured).

No documentation changes: no doc describes this UI state. The "billed cost may
remain pending" line in `docs/deployment/seqera-cost-labels.md` is about Cost
Explorer tag attribution, not this row, so it still reads correctly.

## Checklist*


diegomayorga-dept (approved):
LGTM

PR #935, opened 2026-09-15
fix(egv-305): add manual, per-lab AI failure analysis with config validation and structured logging

diegomayorga-dept:
## Type of Change*
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description

Adds AI-assisted failure classification for failed HealthOmics/Seqera runs, gated behind a per-lab toggle and always technician-triggered — no run is ever queued for analysis automatically.

**Root fix:** all three LLM providers (Bedrock/OpenAI/Anthropic) used to catch every exception and return the same `AMBIGUOUS_FALLBACK` value used for a genuine "the model couldn't decide" result — so a provider error (bad model ID, rejected key, rate limit) was silently indistinguishable from ambiguity. Providers now return a discriminated `ClassificationOutcome` (`{outcome:'classified'}` | `{outcome:'failed', error}`), consumed consistently by the classification consumer, the front-end error map, and structured logs.

**Config validation:** a live probe (`LLMClassificationService.validateConfig`) is wired into laboratory save (`update-laboratory.lambda.ts`), rejecting a bad provider/model/key *before* it's stored, rather than only discovering it on the next failed run.

**Manual-trigger only, by design decision made mid-implementation:** a per-lab `FailureAnalysisEnabled` field (Laboratory schema) is a master switch admins/lab managers use to enable or disable AI error analysis for their lab. When on, a "Run/Re-run AI analysis" button appears on the run detail page (only for failed runs where the lab has a provider configured); when off, the button is hidden for everyone and the backend rejects a bypassed request synchronously (`EG-338`). There is no automatic path — the pre-existing auto-publish-on-failure call site (`process-update-laboratory-run.lambda.ts`) is removed; its function (`safePublishForClassification`) is kept, exported, and unreferenced for a one-line re-enable if product wants it back later.

**Request flow:** the button posts to a new endpoint (`request-laboratory-run-failure-analysis.lambda.ts`), which validates synchronously and enqueues to the existing SQS classification queue with `Trigger: 'Manual'` — the consumer uses this to bypass its idempotency guard, so a re-run after fixing a bad config actually re-runs. `AnalysisStatus`/`AnalysisErrorCode`/`AnalysisErrorMessage`/`AnalysisRequestedAt` (new optional `LaboratoryRun` fields) are polled by the front end (3s interval, 90s timeout, resumes on page load if a prior request left a run stranded in-flight).

**Observability:** the classification consumer's raw `console.log('EVENT: ...')` dump and scattered `console.warn` calls are replaced with one structured JSON line per execution (`logAnalysisEvent`), including on the previously-unhandled exception path — no log excerpt, failure text, or API key in the payload.

**Front-end:** a code-to-copy error map (`analysisErrorMessage`) used by both the synchronous rejection and the async polled failure; the "Failure analysis" card was moved above the Run Details/File Manager tabs, next to Failed Tasks, so it's visible without switching tabs; the Lab Settings toggle collapsed to a two-state Enabled/Disabled badge (was three-state before the manual-only pivot); the pre-existing "Analyse run logs on failure" (CloudWatch log enrichment) toggle is hidden from the settings form to keep the surface simple, per the same product decision — the underlying field/behavior is untouched, a lab that already had it on keeps that behavior.

**Out of scope / follow-up:**
- Bedrock IAM grants (both the manual probe and the consumer) cover `foundation-model/*` only, not `inference-profile/*` — a real gap for cross-region inference-profile model IDs (e.g. `us.anthropic.claude-*`). Flagged, not fixed — needs a decision on which model IDs are actually configured before widening IAM scope.
- `create-laboratory` has no live config probe (only `update-laboratory` does) — a new lab can be created with a bad model ID and only discovers it on the first failed run.
- A pre-existing duplicate `EG-333` error code was noticed in `HttpError.ts` during this work; unrelated to this PR, not touched.

## Testing*

**Automated (all passing as of the latest commit):**
- shared-lib: 23/23 suites, 160/160 tests
- back-end: 124/124 suites, 993/993 tests
- front-end: 18/18 suites, 149/149 tests

New/updated coverage includes the provider outcome discrimination, the config-validation probes, the manual-trigger endpoint's authorization/gating/EG-337/EG-338 paths, the consumer's trigger/toggle matrix, the front-end polling store (including that intervals genuinely stop firing, not just that state is cleared), and the Lab Settings save-payload regression this session found (see below).

**E2E:** two new Playwright specs — button visibility for a lab technician, and the synchronous invalid-model-ID rejection for an org admin (the only role with Lab Settings access; the original scope had this on the wrong role and was corrected). Confirmed discoverable via `playwright test --list`; **not executed against a live `quality` environment** — no deployment was reachable during authoring. The funded-provider success path is documented as manual QA steps rather than automated, since it costs real tokens against a real LLM.

**Manual (done live during this session, by Diego, against a local dev server + a real dept-dev AWS account):** this surfaced and led to fixing three real bugs that automated tests alone hadn't caught:
- The Lab Settings toggle appeared to save successfully but never actually persisted — `update-laboratory.lambda.ts` builds an explicit field list for the DynamoDB write, and `FailureAnalysisEnabled` had never been added to it (true since the field's introduction, not a regression from a later commit).
- Both the manual endpoint and the consumer wrote `AnalysisErrorCode: undefined` on their success paths, which this repo's `marshall()` call throws on with no `removeUndefinedValues` — every successful classification and every trigger request would have thrown in production.
- Lab Settings kept its own local component state and never wrote a fresh save back into the shared Pinia lab store, so other pages (the run detail page's button gating) kept reading stale data until a hard reload.

**Not yet confirmed:** the full happy path (click → queue → consumer → real Bedrock call → success) was blocked in this session by two environment issues unrelated to this branch's code — an expired AWS SSO session, then CloudFormation stack drift (two DynamoDB tables from unrelated features already exist outside this stack's state, blocking any deploy to `dept-dev`). Needs a clean deploy to confirm end-to-end.

## Impact

- **Backend:** one new Lambda + HTTP route (`POST .../request-laboratory-run-failure-analysis`), two new IAM grants (the new endpoint's DynamoDB/SQS access; `bedrock:InvokeModel` on `update-laboratory` for the config probe — both scoped no more broadly than needed), two new typed errors (`EG-337`, `EG-338`).
- **Schema:** `LaboratoryRun` gains 4 optional fields, `Laboratory` gains 1 — all additive, `!== false` default-on semantics, no data migration for existing records.
- **Behavior change (deliberate):** AI failure analysis no longer runs automatically on any failed run. This is a live-in-development auto-trigger being removed by product decision, not a regression — see Description.
- **New network calls:** one live provider probe (Bedrock/OpenAI/Anthropic) on laboratory save, only when the provider/model/key actually changed.
- **Front-end:** no new dependencies.
- **No new npm dependencies** anywhere in the diff.

## Additional Information

This PR's shape changed mid-implementation: the original design had automatic analysis on by default with the toggle only affecting *automatic vs. manual*. A Slack decision between Diego and Stanislav (recorded and reasoned through in the branch's commit history) changed this to manual-trigger-only before the branch was finalized — the final state described above reflects that decision, not the original plan.

One pre-existing ESLint warning (`import/order`) shows up in back-end test output on `process-update-laboratory-run.lambda.test.ts`, a file this PR modifies (added an import for the now-exported `safePublishForClassification`). It wasn't independently confirmed against the base commit whether this warning predates the PR — flagging for the reviewer's awareness rather than asserting either way.

## Checklist*
- [x] No new errors or warnings have been introduced. *(0 lint errors across all three packages; one pre-existing-looking `import/order` warning noted above, not independently confirmed pre-existing — see Additional Information)*
- [ ] Documentation has been updated accordingly. *(Updated four documents that live outside this repo — CLAUDE.md, OBSERVABILITY.md, TESTING.md, and the canonical ERROR_HANDLING.md at the workspace root — not visible in this diff and not part of this PR)*

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)


ignacionistal (commented):
## Review — request changes

Security looks solid (authz, lab/run binding, secret-safe logging, async re-validation). Blocking on status-lifecycle bugs that can leave the UI polling forever.

### High

**1. Disabled mid-flight leaves non-terminal status**  
`process-classify-laboratory-run-failure.lambda.ts` (~137–148)

When the consumer skips because `FailureAnalysisEnabled === false`, it logs and returns without writing a terminal `AnalysisStatus`. A run already set to `Queued`/`Running` by the request handler never completes, so the front-end keeps polling until the 90s timeout.

**Suggested fix:** on the skip path, update the run to a terminal status (e.g. `Failed` with a clear code, or clear `AnalysisStatus`) before returning.

**2. Unhandled exceptions leave status on `Running`**  
`process-classify-laboratory-run-failure.lambda.ts` (~201–212)

For manual analysis, status is set to `Running` before classification, but the outer `catch` only logs and rethrows. Timeouts, DynamoDB errors, or other exceptions leave the run stuck on `Running` with no terminal outcome.

**Suggested fix:** in the `catch`, best-effort write `AnalysisStatus: 'Failed'` (and a safe error code/message) before rethrowing.

### Medium

**3. SQS failure after `Queued` write**  
`request-laboratory-run-failure-analysis.lambda.ts` (~67–85)

`AnalysisStatus: Queued` is persisted before `sqsService.sendMessage`. If SQS fails after the DynamoDB write, the run stays queued with no consumer message and polling never reaches a terminal state.

**Suggested fix:** send to SQS first, then write `Queued`; or on SQS failure, compensate by reverting/clearing status (or writing `Failed`).

**4. No in-flight guard on the request endpoint**  
`request-laboratory-run-failure-analysis.lambda.ts` (~49–77)

The endpoint does not reject when `AnalysisStatus` is already `Queued` or `Running`, so repeated clicks enqueue parallel classification jobs for the same run.

**Suggested fix:** reject synchronously if status is already in-flight (or make enqueue idempotent per run).

**5. Polling starts without optimistic UI lock**  
`packages/front-end/src/app/stores/run.ts` (~235–272)

After a successful request, polling starts on a 3s interval without optimistically setting `AnalysisStatus` to `Queued` (and without an immediate poll). The button can stay enabled for a short window and allow duplicate submissions.

**Suggested fix:** optimistically set local status to `Queued` (or disable the button immediately) and/or poll once right away.

### Non-blocking notes (author already flagged)

- Bedrock IAM covers `foundation-model/*` only, not `inference-profile/*`
- Live LLM config probe on `update-laboratory` but not `create-laboratory`
- E2E / full happy path not yet confirmed against a deployed environment

Happy to approve once the two highs (and ideally the SQS / in-flight mediums) are addressed.

diegomayorga-dept:
@ignacionistal — all five addressed. Two deviate from the suggested fix, with reasoning below.

### High

**1. Disabled mid-flight leaves non-terminal status** — the skip path now writes a terminal `Failed` with `AnalysisErrorCode: ANALYSIS_DISABLED` before returning, gated on `isManual`. An automatic trigger has no poller and no prior status write, so writing a terminal status there would only add noise.

**2. Unhandled exceptions leave status on `Running`** — the outer `catch` now best-effort writes `Failed` / `UNEXPECTED_ERROR` before rethrowing. `isManual` moved above the `try` so the catch can see it, and a failure to write the status is swallowed and logged so it cannot mask the original error.

Both go through one shared `markAnalysisFailed` helper.

### Medium

**3. SQS failure after `Queued` write** — kept write-then-send and added a compensating write (`ANALYSIS_NOT_QUEUED`) rather than reordering. Sending first and then writing `Queued` trades a stuck run for a clobbered one: the consumer can write `Running` before the handler gets to its `Queued` write, and the handler would then overwrite it.

**4. No in-flight guard** — added `LaboratoryFailureAnalysisInProgressError` (EG-339, 409). It is gated on a staleness window equal to the classification queue's 5-minute visibility timeout rather than rejecting unconditionally: a hard-killed consumer invocation writes no terminal status at all, so an unconditional guard would lock that run out of analysis permanently. An in-flight status carrying no parseable `AnalysisRequestedAt` is treated as stale for the same reason.

**5. Polling starts without optimistic UI lock** — the status is now set to `Queued` *before* the request is sent, not after it returns; the button's disabled state derives from `AnalysisStatus`, so the round-trip itself was the window that allowed a double-submit. It is restored to the previous value if the request is rejected. Polling also fires once immediately instead of waiting a full interval.

### Also in this round

- Front-end copy added for the three new codes (`ANALYSIS_DISABLED`, `ANALYSIS_NOT_QUEUED`, `UNEXPECTED_ERROR`). No schema change or API diff — `AnalysisErrorCode` is already `z.string()`.
- EG-339 documented in `ERROR_HANDLING.md`, along with EG-338, which shipped earlier in this PR but never made it into the reference table.

### Verification

Both lambdas were reverted to the pre-fix commit and the suite re-run: exactly the 5 new behavioural tests fail without the fixes, confirming they test the fix rather than passing either way.

| package | suites | tests |
|---|---|---|
| shared-lib | 23 | 160 |
| back-end | 124 | 1004 (was 993) |
| front-end | 18 | 152 (was 149) |

Lint clean across all three. The pre-existing `import/order` warning on `process-update-laboratory-run.lambda.test.ts` is untouched.

The three non-blocking notes (Bedrock `inference-profile/*` IAM, no config probe on `create-laboratory`, unverified E2E happy path) remain as flagged — still out of scope for this PR.

ignacionistal (approved):

PR #936, opened 2026-09-15
fix: skip full data collections s3 scan on page load

ignacionistal:
## Type of Change*
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
- Stop kicking off a full lab-bucket listing (and per-object DynamoDB linkage checks) whenever Data Collections mounts. Samples and Sequence Collections now load from DynamoDB only.
- Cache unlinked-file scan results in a session Pinia store (5-minute TTL). Standard tab navigation reuses the cache; Files shows it immediately and refreshes in the background when stale. Rescan, import, and sample creation still force a fresh scan.
- Batch DynamoDB linkage lookups with `BatchGetItem` so a scan that does run is much cheaper than one `GetItem` per S3 object.

## Testing*
- [ ] Open Data Collections on a lab with a large bucket (~thousands of files). Samples should appear without a long S3-scan wait; Files badge can show `—` until a scan has run.
- [ ] Open the Files tab once and confirm the scan runs. Leave Data Collections, come back: list and count should appear immediately with no new scan.
- [ ] Click **Rescan bucket** and confirm a fresh scan runs.
- [ ] Build a sample from unlinked files (or import), then reopen Files: previously linked files should no longer appear.
- [ ] Lab with no default S3 bucket still shows the settings banner and does not scan.

## Impact
This removes the full lab-bucket listing from Data Collections page load. Samples and Sequence Collections should render from DynamoDB without waiting on S3, which is the main win for labs with large buckets (~thousands of files).

The Files tab still scans S3, but only on first open in a session, after a 5-minute cache expires, or when the user clicks Rescan. Switching away from Data Collections and back reuses the in-memory cache, so standard navigation no longer remounts into another full walk. When a scan does run, linkage checks use DynamoDB BatchGetItem instead of one GetItem per object, so that path is cheaper too.

Behaviour changes: the Files badge shows — until a scan has run in this session (it no longer shows 0 while pretending the bucket was listed). Import, sample creation, and Rescan still invalidate/refresh so linked files do not linger. Cache lives in Pinia for the browser session only — a full reload will scan again the next time Files is opened.

No new AWS resources, no new npm dependencies, and no API contract changes.

## Checklist*


diegomayorga-dept:
# PR Review: fix: skip full data collections s3 scan on page load (#936)

## Summary

- **0 blockers, 7 major, 12 minor**
- Security and Breaking-changes came back clean.
- Top concerns:
  1. **In-flight scans are never scoped to a lab or cancelled** — three major findings (lab-switch cache poisoning, duplicate concurrent scans, invalidation undone by a late response) all trace to one root: `fetchUnlinkedFiles` reads `props.labId` *after* the await and nothing tracks request generation.
  2. **`force: true` is passed at every call site**, making the in-flight de-dupe guard dead code — so the expensive S3 walk this PR exists to avoid can still run 2–3× concurrently.
  3. **The BatchGetItem chunking loop is never tested above 2 refs** — the >100-key path is the whole point of the back-end change.

---

## Correctness (4)

### major — `packages/front-end/src/app/components/EGDataCollectionsPage.vue:179`

`fetchUnlinkedFiles` reads `props.labId` **after** the await, so a scan started for lab A is written into the cache under whatever lab is current when it resolves. The component isn't remounted on lab change (there's an explicit `watch` on `props.labId`), so switching labs mid-scan stores lab A's `Contents`/`S3Bucket`/`ResolvedPrefix` under lab B's key — and the user can then select those objects and build samples from another lab's files. The error path's `clearUnlinkedFiles()` has the same post-await read and can clear the wrong lab's cache.

**Suggestion:** Capture `const labId = props.labId` before the request, pass it to `requestUnlinkedBucketObjects` / `setUnlinkedScan` / `clearUnlinkedScan`, and discard the response if `labId !== props.labId` on resolve.

### major — `packages/front-end/src/app/components/EGDataCollectionsPage.vue:210`

The de-dupe guard `if (unlinkedFilesRequest && !opts.force)` is unreachable — `ensureUnlinkedFilesLoaded` passes `{force: true}` on both the no-cache and stale branches, and `@rescan` does too. Toggling Files → Samples → Files during a 25k-key scan fires a second full bucket scan, then a third. Compounding it: `unlinkedFilesRequest = null` in the `finally` of whichever finishes first, and the shared `uiStore` `'dataCollectionsList'` flag is cleared by the first completion, so the Rescan spinner stops while a scan is still running.

**Suggestion:** Have `ensureUnlinkedFilesLoaded` call `loadUnlinkedFiles()` *without* force and let it own the decision; for a forced rescan, abort/ignore the outstanding request rather than running two in parallel.

### major — `packages/front-end/src/app/components/EGDataCollectionsPage.vue:521`

Cache invalidation can be undone by an in-flight scan. On the Files tab with a stale cache, a background `void loadUnlinkedFiles({force:true})` runs while the cached list is shown. If the user completes Build Sample / Build Samples From Regex / Import in that window, the handler calls `clearUnlinkedScan(labId)` — then the outstanding scan resolves and calls `setUnlinkedScan` with the **pre-link snapshot**, resurrecting just-linked files as unlinked with a fresh `scannedAt`, so the bad entry survives the full 5-minute TTL. This is the exact failure the PR's own test plan step 4 is meant to catch.

**Suggestion:** Track a per-lab request generation (bumped by `clearUnlinkedScan`) and discard responses whose generation changed.

### minor — `packages/back-end/src/app/services/easy-genomics/laboratory-sample-service.ts:635`

The `UnprocessedKeys` retry loop re-sends immediately with no backoff. `UnprocessedKeys` is returned *because* of throughput throttling, so three back-to-back retries likely return the same keys, and the per-key `GetItem` fallback then fires with no delay either. All keys share one `LaboratoryId` partition and `ConsistentRead` doubles capacity cost — a hot-partition burst on a large scan.

**Suggestion:** Exponential backoff with jitter between attempts, per the AWS `UnprocessedKeys` guidance.

---

## Conventions (3)

### major — `packages/front-end/src/app/components/EGDataCollectionsPage.vue:175`

Introduces `useDataCollectionsStore` but leaves the repository call in the component: `fetchUnlinkedFiles()` calls `$api.dataCollections.requestUnlinkedBucketObjects(...)` directly, then pushes the result into the store. This inverts the Pages → Stores → Composables → Repository rule — compare `stores/labs.ts` `loadLab()`, which owns the `$api` call, the try/catch, and the `useToastStore().error()`.

**Suggestion:** Move the fetch into `dataCollections.ts` as `loadUnlinkedScan(labId, { force })` owning the call, TTL check, and error toast. This would also make the three correctness findings above structurally easier to fix — one place owns lab scoping and in-flight state.

### minor — `packages/front-end/src/app/stores/dataCollections.ts:4`

`UnlinkedFile` re-declares a shape that already exists in shared-lib as the element type of `UnlinkedBucketObjectsResponse['Contents']`. The two can drift.

**Suggestion:** `type UnlinkedFile = NonNullable<UnlinkedBucketObjectsResponse['Contents']>[number];`

### minor — `packages/front-end/src/app/components/EGDataCollectionsPage.vue:189`

`applyCachedUnlinkedScan` must be manually re-called at six call sites to keep local refs in sync with the store — shared logic living in a component.

**Suggestion:** Expose the scan as a computed off the store (no mirroring needed), or extract to a `useUnlinkedFilesScan.ts` composable.

---

## Tests (3)

### major — `packages/back-end/test/app/services/easy-genomics/laboratory-sample-service.test.ts:361`

The `BATCH_GET_ITEM_LIMIT = 100` chunking loop is never exercised — both new tests pass 1–2 refs. Chunking is the core of this PR: the scan now passes every object in the bucket in one call, and DynamoDB hard-rejects >100 keys. A slice bug either throws on any lab with >100 files, or silently drops refs so already-linked files report as unlinked.

**Suggestion:** Test with ~250 refs; assert `batchGetItem` called 3×, no call exceeds 100 keys, requested keys equal the input refs, and all chunks merge into the returned Map.

### major — `packages/front-end/src/app/components/EGDataCollectionsPage.vue:196`

The largest change in the PR (+108/−36) is untested. The repo has no `.vue` component tests, so a component test is correctly out of scope — but the new cache branching (cache/stale/force/in-flight) is plain testable logic, and the repo's established pattern for data-collections logic is a pure helper in `src/app/utils/` with a matching test (`data-collections-filters.test.ts`, `data-collections-selection.test.ts`). The gap is *where the logic was put*.

**Suggestion:** Extract `resolveUnlinkedScanAction({ hasCache, isStale, force, requestInFlight })` into `src/app/utils/` and cover its branches.

### minor — `packages/back-end/test/app/services/easy-genomics/laboratory-sample-service.test.ts:390`

`UnprocessedKeys` is only tested all-or-nothing (always unprocessed → exhausts retries → `getItem` fallback). The **partial**-retry merge — attempt 2 succeeds for some keys while narrowing `keys` — is untested, and that merge is where retry loops typically drop results.

---

## Simplification (6)

### minor — `packages/front-end/src/app/components/EGDataCollectionsPage.vue:196` and `:88`

`loadUnlinkedFiles({force})` duplicates the freshness logic `ensureUnlinkedFilesLoaded` already does, and since every caller forces, both `applyCachedUnlinkedScan()`'s early-return and the entire `unlinkedFilesRequest` bookkeeping are dead code. Same root cause as the correctness finding at `:210` — fixing one fixes both.

### minor — `packages/back-end/src/app/services/easy-genomics/laboratory-sample-service.ts:630`

Third hand-rolled chunk-and-batch-get loop against the same table — `laboratory-data-tagging-service.ts` has the same shape twice (~715, ~756). The DRY rule says extract before the third copy, and only this copy handles `UnprocessedKeys`, so behavior now silently differs across the three.

**Suggestion:** Add `protected batchGetAll(tableName, keys, { consistentRead })` to `DynamoDBService`, owning chunking + UnprocessedKeys, and have all three call sites use it.

### minor — `packages/back-end/src/app/services/easy-genomics/laboratory-sample-service.ts:657`

Two overlapping recovery mechanisms for one failure: three BatchGetItem re-attempts *and* a per-key GetItem fallback. Either alone covers throttled keys.

### minor — `packages/front-end/src/app/components/EGDataCollectionsPage.vue:328`

`clearUnlinkedScan(labId)` + `applyCachedUnlinkedScan()` is inlined at 328, 521 and 537 — that's exactly what the existing local `clearUnlinkedFiles()` already does.

### minor — `packages/front-end/src/app/stores/dataCollections.ts:10`

`UnlinkedScan.isTruncated` is written by `setUnlinkedScan` but never read anywhere.

---

## PR hygiene (3)

### major — PR title

No ticket reference. Easy Genomics uses a ticket-prefix convention, not conventional commits.

**Suggestion:** `EGV-XXX Skip full data collections S3 scan on page load`.

### minor — PR body

No link to the ticket/issue — no traceability back to the requirement.

### minor — PR body

"Documentation has been updated" and "new tests added as necessary" are both checked, but there's no `TESTING.md` update and no Playwright spec in the files list, while every Testing checkbox is unchecked.

---

## Clean dimensions

**Security (0)** — auth path untouched (`assertSequenceCollectionsAccess` still gates the handler); BatchGetItem keys are `marshall`'d with the authorized `LaboratoryId` as partition key, so results are inherently lab-scoped; the new store is *not* `persist: true` and is wired into `resetStores()`, so nothing hits localStorage and the cache is wiped on logout.

**Breaking changes (0)** — `getSampleIdsForFileRefs` signature and contract unchanged (2 callers, both fine); `dynamodb:BatchGetItem` is already granted on the tagging table at `easy-genomics-nested-stack.ts:2089-2096`; `fileCount` prop widened `number → number | null` with its only consumer updated in the same diff.

---

**Cross-cutting note:** Correctness `:179`, `:210`, `:521` and Conventions `:175` are all consequences of the same structural choice — the component, not the store, owns the fetch, the lab scoping, and the in-flight state. Moving `fetchUnlinkedFiles` into the store as a lab-keyed action with a generation counter would address all four at once.

diegomayorga-dept (approved):
## PR Review

Defers the full S3 unlinked-file scan off Data Collections page load (Files tab only, with a 5-minute Pinia session cache) and replaces per-object DynamoDB `GetItem` linkage checks with chunked `BatchGetItem`.

No critical issues found. LGTM.

✅ (approved)

PR #937, opened 2026-09-16
chore: bump posthog-js to address outdated sdk alert

ignacionistal:
## Title*
Bump posthog-js to resolve PostHog outdated Web SDK alert

## Type of Change*
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
PostHog SDK Health flagged Web SDK `1.386.7` as outdated. This updates `posthog-js` from `^1.386.7` to `^1.433.5` and refreshes only the related lockfile entries (`@posthog/*`, `preact`, `web-vitals`).

Also pins new SDK defaults that would otherwise expand capture surfaces in a healthcare context:
- `capture_performance: false`
- `disableDeviceModel: true`
- `save_campaign_params: false`

Lockfile was re-resolved with the repo’s pinned `pnpm@9.15.0` after rebase onto `development` to avoid unintended amplify/react-native/`@babel/core` duplicate-tree drift from the initial install.

## Testing*
- [x] `pnpm install --frozen-lockfile` succeeds
- [x] Front-end unit tests + eslint pass (`201` tests)
- [ ] Full `pnpm build-front-end` blocked locally by expired AWS token at `nuxt-load-settings` (unrelated)
- [ ] After deploy, confirm PostHog SDK Health shows `1.433.x` and the outdated alert clears

## Impact
Front-end dependency bump plus a small analytics init hardening. Lockfile changes are limited to PostHog and its direct transitive deps; no amplify/react-native babel-core split.

## Additional Information
- `sanitize_properties` → `before_send` migration left for a follow-up (deprecated in both old and new SDK).
- CI still does not build/test PRs on push to feature branches (pre-existing).

## Checklist*


diegomayorga-dept (commented):
## PR Review (deep)

**Size:** +1060/-122 — 1059 of which are `pnpm-lock.yaml`.

**Summary:** 2 major, 0 blockers, plus some informational notes on the SDK delta. The `posthog-js` bump itself is API-compatible and carries no CVE exposure. The concerns are about the lockfile and about nothing in CI validating it.

---

### major — `pnpm-lock.yaml`: the bump re-resolved `@babel/core` downward and split the tree

Verified by direct comparison of the two lockfiles:

| | `react-native@0.75.4(@babel/core@7.29.7)` | `…(@babel/core@7.26.10)` |
|---|---|---|
| `origin/development` | **106** | 0 |
| this branch | **14** | **95** |

`development` resolves the whole aws-amplify / react-native subtree against a single `@babel/core@7.29.7`. This branch replaces most of it with `7.26.10` and leaves 14 entries on `7.29.7` — so that subtree is now instantiated **twice**, under two different babel cores, where it previously resolved once.

Only one specifier actually changed in the PR (`posthog-js`), so this looks like unintended install drift rather than a decision.

Functional runtime risk is low — `react-native` is an unused optional transitive of amplify v5 in a web app — but it inflates a ~40-line change into 1059 lines, duplicates an install subtree, and will conflict with every other open branch that touches the lockfile.

**Suggestion:** rebase on `development` and re-run `pnpm install` with the pinned `pnpm@9.15.0` from `packageManager` (a different local pnpm/node version, or a stale store, is the likely cause). Then confirm the diff only touches `posthog-js`, `@posthog/*`, `preact`, and `web-vitals` entries.

### major — nothing in CI builds or tests this PR

`gh pr checks 937` shows exactly two checks: `API schema diff check` and `Validate PR title`. Build/test/deploy workflows run only `on: push` to `development` / `staging`. So a lockfile that breaks install or the front-end build would land on `development` and flow straight into the `quality` deploy.

This is a pre-existing repo-wide gap rather than something this PR introduced, but it's what makes the point above worth resolving before merge rather than after.

**Suggestion:** at minimum run `pnpm install --frozen-lockfile && pnpm build-front-end` on the rebased branch and note it in the Testing section.

---

### Informational — posthog-js 1.386.7 → 1.433.5 behavioural delta

Researched against the upstream changelog, the tagged `posthog-core.ts` source, and the published type declarations for both versions.

**Compatible:** `import('posthog-js')` still yields `.default`. `identify`'s first arg narrowed from optional to required, but `useAnalytics.ts:315` guards with `if (hashed)`, so it's unaffected. `sanitize_properties` still runs through the single `calculateEventProperties()` funnel, so it covers all event types including `$web_vitals` / `$exception` / `$dead_click`. No advisories in range — the only posthog-js records are CVE-2023-32325 (fixed in 1.57.2) and MAL-2025-191402 (only 1.297.3).

Four items worth a deliberate decision given the healthcare context. **None are defects in this PR** — just new default behaviour arriving with the bump:

1. **`$device_model` is now captured unconditionally** via `navigator.userAgentData.getHighEntropyValues(['model'])` on Android Chromium. `disableDeviceModel` defaults to `false`. New fingerprinting surface, not gated by the consent config.
2. **Web-vitals attribution is on by default** for INP/LCP — `$web_vitals` events now carry DOM target-element identifiers. Only fires if `capture_performance` is enabled in the PostHog **project settings**; the init block leaves that to remote config rather than pinning it locally, unlike every other capture flag there.
3. **New Meta `_fbc` / `_fbp` cookie reads**, gated by `save_campaign_params` (defaults `true`). Practically inert with no Meta pixel present.
4. **Bundle grows ~202 KB → ~316 KB raw** (uncompressed `dist/module.js`). Dynamically imported behind the consent gates, so no cost for users who never opt in.

Adding `capture_performance: false`, `disableDeviceModel: true`, and `save_campaign_params: false` would match the locally-pinned posture the rest of that init block already takes.

Separately and unrelated to this PR: `sanitize_properties` is **deprecated in both the old and the new version** and logs an error on every captured event. Migrating to `before_send` deserves its own ticket.

---

### minor — PR description

The checklist marks "Documentation has been updated accordingly" and "new tests added as necessary" as done, but the diff contains no doc or test changes. The description also says "Front-end only dependency bump" — true of `package.json`, but the lockfile change affects root/back-end importer resolution too.

---

☑️ Issues found — the lockfile churn is worth cleaning up before merge, especially given no CI gate covers it.

ignacionistal:
@diegomayorga-dept  Thanks for the deep review — addressed the actionable items in the follow-up commit(s):

- Rebased onto `development` and re-resolved `pnpm-lock.yaml` with the pinned `pnpm@9.15.0`, so the lockfile diff now only touches `posthog-js`, `@posthog/*`, `preact`, and `web-vitals` (no amplify/react-native `@babel/core` split).
- Pinned the new SDK defaults locally: `capture_performance: false`, `disableDeviceModel: true`, `save_campaign_params: false`.
- Verified `pnpm install --frozen-lockfile`; front-end unit tests + eslint pass. Full `pnpm build-front-end` couldn’t complete locally because AWS credentials expired at `nuxt-load-settings` (unrelated to this change).

**Left alone (intentionally):**

1. **`sanitize_properties` → `before_send` migration**  
   Agreed this is worth doing, but it’s deprecated in *both* the old and new SDK and isn’t introduced by this bump. Migrating the callback shape/behavior deserves its own ticket so we can update + regression-test URL sanitization without bundling it into a dependency bump.

2. **Adding a CI build/test gate on PRs**  
   Confirmed this is a pre-existing repo-wide gap (`on: push` only to `development`/`staging`), not something this PR introduced. Happy to help with a separate infra/CI change, but it’s out of scope for this SDK bump.

3. **Bundle size increase (~202 KB → ~316 KB raw)**  
   Accepted as an upstream cost of the newer SDK. Still dynamically imported behind consent gates, so no load cost for users who never opt in.

diegomayorga-dept (approved):

PR #938, opened 2026-09-16
fix: workflow runs retention policy deletion enabled

ignacionistal:
## Title*
Enable hard-delete of expired laboratory data in retention sweep

## Type of Change*
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
The daily `process-expired-laboratory-data` Lambda completes the run-retention cascade by deleting eligible S3 objects and tagging rows after referencing runs expire. It was deployed with `DRY_RUN: 'true'`, so it only logged what it would delete and never performed hard deletes.

This change sets `DRY_RUN` to `'false'` so scheduled retention cycles actually purge expired input files (and their tagging rows) once run TTL has cleared usages. Permanent-tagged and never-used orphan files remain skipped.

## Testing*
- [ ] Confirm after deploy that the Lambda env var `DRY_RUN` is `false`
- [ ] Review CloudWatch logs for `process-expired-laboratory-data` on the next 04:00 run: expect real deletes (no `[dry-run] Would delete` for eligible files)
- [ ] Spot-check a lab with expired runs: eligible input files are removed; Permanent-tagged files are retained
- [ ] Existing unit tests already cover dry-run vs live delete behaviour (`process-expired-laboratory-data.lambda.test.ts`); no new tests required for this env flip

## Impact
- Behaviour change: expired, eligible laboratory input files will be hard-deleted on the daily sweep after run records TTL out (typically within ~48 hours of expiry, then on the next daily job).
- No new dependencies.
- Ops can temporarily set `DRY_RUN=true` on the Lambda to audit eligibility without deleting.

## Additional Information
- Does not make deletion synchronous on retention-policy save; run rows still rely on DynamoDB TTL, and file cleanup remains on the daily schedule after cascade bookkeeping.
- UI copy clarifying the up-to-~48h removal delay is a follow-up (not in this commit).

## Checklist*


diegomayorga-dept (commented):
## Deep review — 6 blockers, 7 major, 3 minor

The three-line diff is correct in isolation, and the intent (completing a retention cascade that shipped inert) is right. The concerns are all about the code this flag arms — which was written and merged while it could not do damage.

### Blockers

**1. Eligibility is inferred from absent state, not proven expiry** — `process-expired-laboratory-data.lambda.ts:131`
The test is `hasWorkflowTag && !hasUsages && !isPermanent`; there is no expiry-timestamp check anywhere in the loop. "No usages recorded" is indistinguishable from "usages expired". The same state is produced by a partially-failed tagging write: in `associate-laboratory-run-workflow-tagging.ts` the workflow tag is applied (~line 51) *before* `recordLaboratoryRunInputUsage` (~line 76), and the block is wrapped in a try/catch that swallows the error. A transient DynamoDB failure between the two leaves an **actively running** job's inputs classified as eligible.
→ Require a positive expiry marker (`UsagesClearedAt`/`LastRunExpiredAt`) written by the stream subscriber. At minimum reverse the write order so partial failure fails safe.

**2. Manual run deletion cascades to file destruction** — `process-laboratory-run-stream.lambda.ts:76`
`isTtlRemoval` is computed from `principalId === 'dynamodb.amazonaws.com'` and then only logged — the comment says "we treat both the same way — the bookkeeping is identical." `delete-laboratory-run` is permitted to Lab Technicians. A technician tidying the runs list now permanently destroys raw uploaded sequencing data on the next sweep, with nothing expired.
→ Branch on `isTtlRemoval`; only TTL removals should mark files sweep-eligible.

**3. The "Permanent" escape hatch is unreachable from the product** — `process-expired-laboratory-data.lambda.ts:132`
`dataCollections.addTagsToFiles` has zero callers outside the repository module. The only wired path is `addTagsToSamples`, which the back-end explicitly rejects for permanent tags ("Permanent tags are system-managed"). Users have no way to protect anything before the first live sweep.
→ Don't flip until a user-reachable "Mark Permanent" affordance exists, or the sweep honours sample-level permanence.

**4. Deletes are irreversible — no recovery path exists**
Lab buckets are created without versioning in both `s3-construct.ts` and the runtime `S3Service.createBucket`. No delete marker, no soft-delete window, no quarantine. Any single eligibility misjudgement above is permanent.
→ Enable versioning, or soft-delete via object tagging + lifecycle expiry (least invasive), for a ~30-day reversal window.

**5. First live run cannot drain its own backlog in 900s**
Work is fully sequential (labs one at a time, ~6 serial round-trips per file) — roughly 6–9k files per invocation — while `MAX_DELETES_PER_LAB_SWEEP` is never set in CDK and defaults to **25,000 per lab**. On timeout the summary log and `emitEmfMetrics` never execute, so there's no record precisely when it matters; labs are walked in fixed order with no resume cursor, so labs late in the list starve indefinitely.

**6. Hardcoded literal ships live deletes everywhere at once**
dev, quality and prod all flip simultaneously, and every self-hosting customer switches to irreversible hard-delete on their next `build-and-deploy` with no opt-out. The established pattern sits ~8 lines below: `process-sync-run-costs` is gated by `this.props.costExplorerEnabled`, flowing `easy-genomics.yaml` → `ConfigurationSettingsSchema` → stack props. The stated mitigation ("Ops can temporarily set `DRY_RUN=true`") is CDK drift — the next deploy reverts it and silently re-arms deletion.
→ Add `data-retention.hard-delete-enabled` (default `false`) to `ConfigurationSettingsSchema` and thread it as `costExplorerEnabled` is. The schema is `.strict()`, so add the key and example YAML together. Then stage: non-prod first, reconcile `Deleted`/`Errors` against prior `Eligible` counts, then prod with a small `MAX_DELETES_PER_LAB_SWEEP`.

### Major

- **User-facing copy misrepresents the feature.** Every description promises deletion of *run records only* — `EGFormLabDetails.vue:1107` ("0 = never delete run records"), `EGLabView.vue:99`, and the save-confirmation dialog, which enumerates runs only. An Org Admin lowering retention 12 → 1 month sees a dialog about run metadata and irreversibly destroys lab input data. This is wider than the ~48h delay the description defers, and should ship *with* the flip.
- **Live deletes are not auditable** (`:166`). Dry-run logs every candidate; the live path logs only failures and counters — successes emit nothing. With no versioning and no S3 access logging there is no record of what was destroyed. This also invalidates the test step "expect real deletes (no `[dry-run] Would delete`)": absence of that line is equally consistent with the sweep having done nothing.
- **IAM `s3:::*/*` DeleteObject grant is now live** (`easy-genomics-nested-stack.ts:1803`). Previously inert because `DRY_RUN` kept execution off the delete path; application asserts are now the only perimeter. This wildcard is a pre-existing repo-wide pattern (~9 occurrences), not introduced here — but this PR is what arms it. Worth scoping to `arn:aws:s3:::${namePrefix}*/*`.
- **Dangling sample/collection membership** (`:176`). `deleteFileRowAndAssociations` removes the FILE# row and MAP# rows but not sample-file membership rows. Sequence Collections keep listing files that no longer exist; runs launched from them fail on missing inputs.
- **Eventually-consistent read feeds a hard-delete decision** (`laboratory-data-tagging-service.ts:1153`). `listAllFileRowsForLab` omits `ConsistentRead`, unlike `listTags` directly above it which sets it deliberately. A stale read missing a just-written usage entry deletes an in-flight run's input.
- **No test covers the one thing this PR changes.** `DRY_RUN` appears nowhere under `packages/back-end/test/infra/`. The handler treats any non-`'false'` value as dry-run, so a revert or casing typo (`'False'`) silently restores permanent dry-run with green CI — exactly the latent state this PR fixes. The existing lambda tests set `process.env.DRY_RUN` directly; they exercise the handler, not the stack.
- **No pre-merge verification.** All four Testing items are post-deploy and unchecked, on a permanently destructive change, with no stated rollout order. `Documentation has been updated accordingly` is also unchecked while nothing in `docs/` covers the sweep, its eligibility rules, or the `DRY_RUN` lever.

### Minor

- `assertKeyUnderLabPrefix` is untested at the handler level — only the bucket guard is. With live deletes it's the sole guard against a mis-keyed row deleting another lab's object.
- The new comment prescribes a manual console edit as the supported audit procedure, contradicting "All infrastructure is CDK", and the next deploy overwrites it anyway.
- No ticket key in title or branch, so release-scoping by merge branch name has nothing to key on. The body's `Title*` field is clearer than the actual title — worth promoting with a key prepended.

Scope verified clean: the diff is exactly what the description claims, nothing extra or missing.

### Suggested path

Hold until eligibility is a positive expiry test rather than an inferential one, and there's a soft-delete window. Then enable per-environment through config rather than as a literal.

diegomayorga-dept (approved):

PR #945, opened 2026-09-21
fix: added missing functionality to delete workflows created by user

ignacionistal:
## Type of Change*
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
- Adds delete for private HealthOmics workflows created in-app via Create Workflow (`DELETE /aws-healthomics/workflow/delete-private-workflow/{id}`).
- Only org admins and lab managers (the same roles that can create workflows) can delete, and only workflows they created; shared, console-created, and other people’s workflows are rejected.
- Lab workflows table shows a Delete action with confirmation. The API still enforces ownership and blocks delete while a run is using the workflow.

## Testing*
- [ ] As a lab manager, open HealthOmics Workflows and confirm Delete appears on a private workflow you created.
- [ ] Confirm Delete is hidden for shared workflows and for users who cannot create workflows (lab technicians).
- [ ] Delete your own workflow and confirm it disappears from the list.
- [ ] Attempt to delete a private workflow created by someone else and confirm a 403 / error toast.
- [ ] Attempt to delete a workflow that has an in-progress run and confirm it is rejected and the workflow remains.

## Checklist*


diegomayorga-dept:
## PR Review: #945 — delete user-created private HealthOmics workflows

**2 blockers, 7 major, 8 minor.** Top concerns: the ownership signal this PR depends on is never written (and won't compile); a rejected delete still permanently destroys all workflow versions; the UI Delete gate fails open, so it shows on every private workflow.

---

### 🔴 Blockers

**1. `metadata` is not a valid `CreateWorkflowRequest` field** — `packages/back-end/src/app/controllers/aws-healthomics/workflow/create-private-workflow.lambda.ts:290`

`metadata: { createdByUserId: omicsUserId }` does not exist on `CreateWorkflowRequest` in the pinned `@aws-sdk/client-omics@3.1090.0` (the interface has `tags`, no `metadata`). Since `workflowRequest` is a typed spread with no index signature, excess-property checking fires (TS2353) and the back-end build fails. Even if suppressed, the SDK serializer drops unmodeled members, so `createdByUserId` is never persisted — which makes the front-end's `workflow.metadata?.createdByUserId` fallback dead code.

> **Fix:** Drop the `metadata` block. Ownership is already carried by `tags.UserId`, which is what `canDeleteOwnedPrivateWorkflow` actually checks. Remove the matching fallback in `EGLabView.vue`.

**2. Rejected delete still destroys every workflow version** — `packages/back-end/src/app/controllers/aws-healthomics/workflow/delete-private-workflow.lambda.ts:75`

`deletePrivateWorkflow` treats **any** `ConflictException` from the first `DeleteWorkflow` as "this workflow still has versions" and unconditionally runs `deleteWorkflowVersions`, permanently deleting every version. HealthOmics also returns `ConflictException` when a run is using the workflow — the exact case the handler claims to reject. In that path the caller gets a 400 saying *"This workflow cannot be deleted while a run is using it"* and the workflow survives, **after** all non-pinned versions have already been destroyed. Irreversible data loss on a request reported as rejected, and it contradicts the PR body's "blocks delete while a run is using the workflow". The unit test hides this by stubbing `listWorkflowVersions` to return `{ items: [] }`.

> **Fix:** Establish up front whether deletion is safe — list versions first and only enter the version-deletion path when versions exist, and/or inspect the conflict reason to distinguish "has versions" from "in use by a run". Fail closed before deleting anything.

---

### 🟠 Major

**3. Delete gate fails open in the UI** — `packages/front-end/src/app/components/EGLabView.vue:539`

`omicsWorkflowCreatorId` reads `workflow.tags?.UserId || workflow.metadata?.createdByUserId`, but `list-private-workflows.lambda.ts` returns raw `WorkflowListItem`s, and that type has **no `tags` field at all**; `metadata` exists on it but is never populated (see blocker 1). So `createdBy` is always `undefined` and the `if (!createdBy) return true` branch fires — Delete is shown on every private workflow, including other users' and console-created ones. This directly contradicts the PR description.

> **Fix:** Invert the default to `return false` when the creator is unknown, and enrich the list response with the creator id server-side if the UI needs to distinguish.

**4. Delete bypasses the lab-scoped IAM enforcement layer** — `packages/back-end/src/infra/stacks/aws-healthomics-nested-stack.ts:533`

`omics:DeleteWorkflow`/`DeleteWorkflowVersion` are granted straight to the Lambda role on `workflow/*` with no `aws:ResourceTag/LaboratoryId|OrganizationId` conditions, and the handler instantiates a bare `new OmicsService()` instead of `createOmicsServiceForLab(...)`. By contrast, create-private-workflow grants only `sts:AssumeRole`/`sts:TagSession` and assumes the omics-access-role so IAM re-checks principal tags. Net effect: the only thing between an authenticated lab user and deletion of any workflow in the account is the in-handler tag comparison — no IAM backstop.

> **Fix:** Route deletion through `createOmicsServiceForLab(...)` and add the delete actions to the omics-access-role policy under the existing `StringEquals` tag conditions.

**5. Cross-lab blast radius** — `packages/back-end/src/app/controllers/aws-healthomics/workflow/delete-private-workflow.lambda.ts:153`

A private workflow can be granted to many labs via `edit-workflow-access-batch`, and other labs are served from those rows. The ownership check never looks at whether other labs still hold access, so a user in the originating lab can destroy a workflow other labs depend on — and only the caller's access row is removed, leaving dangling records pointing at a deleted workflow.

> **Fix:** Query `laboratory-workflow-access` for all rows on this workflow; reject when other labs are assigned, or remove every access row as part of the operation rather than best-effort.

**6. Enforced roles are broader than documented** — `packages/back-end/src/app/controllers/aws-healthomics/workflow/delete-private-workflow.lambda.ts:36`

The PR says "only org admins and lab managers", and the UI gates on `canEditLabUsers`, but `hasPrivateWorkflowCreateAccess` also accepts `validateLaboratoryTechnicianAccess`. A lab technician who sees no Delete button can delete by calling the endpoint directly. (Ownership tags themselves are not spoofable — written server-side, and no handler exposes `omics:TagResource`.)

> **Fix:** Pick the intended role set and make the server authoritative — drop the technician check, or correct the PR body and UI gate.

**7. Handler carries domain logic** — `packages/back-end/src/app/controllers/aws-healthomics/workflow/delete-private-workflow.lambda.ts:40`

Paginated version deletion plus the delete→purge→retry orchestration live in the controller, totalling 163 lines against the ≤50-line thin-controller rule.

> **Fix:** Move both functions into the service layer (`OmicsService` or a `@BE/services/aws-healthomics/` helper).

**8. Ownership gate has zero test coverage** — `packages/front-end/src/app/components/EGLabView.vue:538`

Including the permissive fallback that is finding 3; the front-end suite has no component tests at all. The repo's pattern for this is a pure helper in `src/app/utils/` with a unit test (see `laboratory-run-ownership.ts`).

**9. Version-deletion paths untested** — `packages/back-end/src/app/controllers/aws-healthomics/workflow/delete-private-workflow.lambda.ts:40`

The pagination loop, the `!version.versionName` skip, and the per-version conflict swallow are all uncovered. The only test touching this path returns a single page with no `nextToken` and a mock that always resolves — a regression dropping pagination would still pass.

---

### 🟡 Minor

- `delete-private-workflow.lambda.ts:40` — versions are deleted while paginating the same list via `startingToken`/`nextToken`; mutating between pages can skip versions, and the retry then reports the misleading "run is using it" 400.
- `delete-private-workflow.lambda.ts:152` — `laboratoryWorkflowAccessService.remove` failure swallowed with a bare `console.warn`; CLAUDE.md permits a swallow only with a comment stating exactly why.
- `delete-private-workflow.lambda.ts:28` — `hasPrivateWorkflowCreateAccess` duplicates the inline permission chain in create-private-workflow; belongs in `auth-utils.ts`, used by both.
- `packages/back-end/src/app/utils/private-workflow-ownership.ts:1` — missing the `*-utils.ts` suffix every sibling uses (`auth-utils.ts`, `omics-shared-workflow-utils.ts`, …).
- `private-workflow-ownership.ts:21` — discriminated union with exactly one possible `reason`; a boolean plus a thrown `UnauthorizedAccessError` says the same thing.
- `packages/front-end/src/app/stores/omicsWorkflows.ts:126` — `deleteWorkflow` calls the API with no try/catch and no `useToastStore().error()`, unlike its sibling `loadWorkflowsForLab`. The error surfaces only because the component happens to catch it.
- `EGLabView.vue:1274` — `omicsWorkflowsActionItems(workflow)` invoked twice per row, rebuilding the array and re-running the predicate. `EGLabView.vue:521` — column list built in three imperative statements to splice one column in; a conditional spread keeps order visible.
- Test gaps: `private-workflow-ownership.test.ts:33` misses `tags: undefined` (console-created workflows), tags without `UserId`, and an OrganizationId mismatch. `omicsWorkflows.test.ts:52` is titled "removes a deleted workflow from every lab list" but loads only one lab, and has no failure-path case.

---

### PR hygiene

- **Type of Change is wrong** — marked Bug fix, but this is a new feature (new handler, new route, new CDK wiring, new UI action). Branch says `feat/`, title says `fix:`.
- **Title typos**, repeated verbatim in the body heading and the sole commit subject: `functionlity` → `functionality`, `ro` → `to`. No ticket key.
- **Generated code mixed with logic** — the OpenAPI yaml/json and `generated.d.ts` share the one commit with hand-written source, against the `chore(generated):` convention.
- Commit has no body explaining why; all five Testing boxes unchecked; "Documentation has been updated" ticked with no doc changes; no E2E coverage for the new happy path.

---

Blockers 1 and 2 are coupled — dropping `metadata` also removes the dead fallback behind the fail-open UI gate.

☑️

diegomayorga-dept:
## PR Review

Adds delete for app-created private HealthOmics workflows — a new `delete-private-workflow` Lambda with tag-based ownership checks, IAM wiring, and a Delete action in the lab workflows table.

### Critical issues

- **N+1 `ListTagsForResource` on every workflow list load** — `private-workflow-list-utils.ts` is called from `list-private-workflows.lambda.ts:108` with an unbounded `Promise.all`: one HealthOmics call per private workflow, on the lab page hot path, for every org admin and lab manager. This repo already needed a bespoke 429 retry loop for `ListShares` (`omics-shared-workflow-utils.ts:44-60`) because HealthOmics throttles aggressively; here throttles are swallowed to `console.warn` and `tags` is left unset, which silently hides the Delete action for workflows the user does own. Both a latency regression and a non-deterministic UI bug. Needs a concurrency cap plus retry, or tags fetched only for the row being acted on.

- **`omics:DeleteWorkflowVersion` is gated on `aws:ResourceTag/LaboratoryId|OrganizationId`** (`aws-healthomics-nested-stack.ts:360-371`). Workflow versions are sub-resources (`workflow/{id}/version/{name}`) and the app never calls `CreateWorkflowVersion`, so nothing tags them — the condition will not match and the call returns `AccessDenied`. `AccessDenied` is neither `ResourceNotFoundException` nor `ConflictException`, so `deleteListedWorkflowVersions` rethrows it raw (`private-workflow-delete-utils.ts:70-79`) and the caller returns an unclassified EG-100 500 instead of a usable error. This branch appears untested against real AWS — verify it against a versioned workflow before merging.

diegomayorga-dept (approved):

PR #946, opened 2026-09-21
fix(egv-310): AI Failure Analysis losing the root cause in noisy HealthOmics logs

diegomayorga-dept:
## Title*
Fix AI Failure Analysis losing the root cause in noisy HealthOmics logs

## Type of Change*
- [x] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [x] UI/UX improvement

## Description

Kelsey reported unhelpful AI failure analyses on HealthOmics runs and sent two
nf-core/viralrecon engine logs. Nextflow on HealthOmics writes a bare pf4j stack
frame for **every task**, leaving up to 96% of an engine log as frames. That noise
filled both the 200-event CloudWatch fetch window and the 4,000-character excerpt
budget, so the root cause never reached the LLM — the Picard run kept
`exit status (2)` but lost the `No registry staging mapping` line that explained it.

Read the commits in order; each stands alone.

**`e9151055` — the fix (start here).**
`extractErrorWindow` now discards noise before choosing the excerpt, via a
`NOISE_RULES` array mirroring the sibling `log-redaction.ts`. `run-log-fetcher`
raises the fetch window from 200 to 1,000 events (no service change — the `limit`
parameter already existed). Measured: 200 and 500 miss the cause, 1,000 captures
it at 76 KB, 2,000+ adds nothing.

**`8553226b` — making the fix reachable.**
`HealthOmicsLogEnrichmentEnabled` gates whether an excerpt is fetched at all, and
it has had no UI control since the Lab Settings surface was trimmed in September —
meaning it was off everywhere and could not be turned on. The toggle is back,
confirmed on enable only (disabling narrows egress), with a provider-aware warning:
OpenAI/Anthropic leave AWS, Bedrock runs under the platform account. `EGDialog`
gains an opt-in `titleTag` prop so this dialog matches the 18px heading other
modals use; every existing dialog is unchanged.

**`79487c7b` — a bug found while testing.**
`AnalysisStatus: 'Queued'` was written before the SQS send and never undone, so a
failed send stranded a run permanently: no consumer picks it up, and the run page
renders "Analysing…" rather than a clickable button. Found a dev run stuck that way
for five days. The revert is exact, so a previous successful analysis survives a
failed re-run request.

**`e5997c38` — closing the same hole from the other side.**
The classification queue had no DLQ, so a record the consumer keeps failing is
retried until it expires and vanishes, producing the identical stranded state with
no trace. Mirrors the existing notification DLQ.

Design and plan: `2026-09-18-ai-failure-analysis-log-extraction` v2.1 in the
AI-tools repo.

## Testing*

Automated:
- `shared-lib` — 167 tests, `log-extraction.ts` at 100% coverage. 5 new tests cover
  noise removal, cause retention through truncation, and a **no-op on clean logs**
  (the noise is workflow-dependent, so a fix that only worked on noisy logs would be
  a regression).
- `back-end` — 1,012 tests / 124 suites, including the infra synth for the DLQ.
  4 new tests assert the run is not left `Queued` when the enqueue throws.
- `front-end` — 220 tests. Neither `pnpm lint` nor the jest lint step covers `.vue`,
  so both changed components were compile-checked with `@vue/compiler-sfc`.
- TDD throughout: each fix had a failing test first.

Manual, against the real logs:

| | `6762520` (Freyja) | `2730770` (Picard) |
|---|---|---|
| excerpt | 4,013 chars | 3,420 chars |
| process + exit status | kept | kept |
| root cause | `Cannot assign requested address` | `No registry staging mapping` + image URI |
| stack frames | none | none |

End-to-end in `dev-demo`: seeded a CloudWatch stream with the Picard log (96%
frames), enabled the toggle, and ran the consumer. `classifiedBy` moved from
`lookup` to `llm` and the excerpt carried the cause.

## Impact

- More CloudWatch data per analysis: ~76 KB instead of ~15 KB, bounded by
  `GetLogEvents`' own 1 MB cap. `DEFAULT_LOG_EXCERPT_CHARS` stays at 4,000.
- The toggle is off by default, so **no lab starts sending logs without opting in**.
- No `ROUTE_SCHEMAS`, schema or dependency changes. The DLQ is a new AWS resource.
- Existing analyses are not regenerated; a run keeps its old text until re-run.

## Additional Information

- **Scope.** This began as the log-extraction fix; the other three commits are
  fallout from trying to test it — the feature could not be reached, then could not
  be retried after a failure. They are separable if you would rather they landed
  apart.
- **Not proven:** that better excerpts produce better *answers*. The excerpt now
  demonstrably contains the cause, but my end-to-end test grafted Kelsey's log onto a
  run whose stored `FailureReason` contradicted it, so the model answered from the
  structured field. Verifying answer quality needs a run whose failure and log agree.
- **Worth checking before release:** whether Kelsey's lab has
  `HealthOmicsLogEnrichmentEnabled` set. If it is off, she never received a log
  excerpt and her original complaint has a different cause.
- The DLQ only exists once deployed — confirm `RedrivePolicy` on the source queue
  after the next dev deploy.

## Checklist*

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ignacionistal (approved):
LGTM

ignacionistal (approved):
LGTM

PR #947, opened 2026-09-22
fix: delete run outputs and sample sheets when a run retention expires

ignacionistal:
## Title*

Delete run outputs and sample sheets when a run's retention expires

## Type of Change*

- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description

The data retention policy only ever deleted a run's **input files**. Workflow outputs under
`results/` and the generated sample sheet were never removed. Because they sit inside the run's
own folder and the run record itself is deleted by TTL, they also become unreachable from the UI
at the same moment — so they are pure storage cost with no remaining way to retrieve them. That
is the opposite of what labs enable retention for.

**Why this could not simply be added to the sweep:** `OutputS3Url` and `SampleSheetS3Url` live
only on the run record, and DynamoDB TTL removes that record *before* the nightly sweep runs. The
sweep has no way to discover where a run's outputs were.

The fix has two halves:

1. **Going forward** — `process-laboratory-run-stream`, which still sees the removed row's old
   image, records the output prefix and sample sheet as a `RUNOUTPUT#` row. The nightly
   `process-expired-laboratory-data` sweep deletes everything under that prefix (including the
   `results/` placeholder object, since S3 folders are only common prefixes) plus the sample
   sheet. This follows the pattern `preserveEmptyFileRow: true` already established for input
   files: the stream handler is the last point that can hand the sweep what it needs.

2. **The existing backlog** — runs that expired under the old behaviour left no `RUNOUTPUT#` row,
   so their data is still orphaned in the bucket. The sweep now runs a reconciliation pass that
   walks each lab's prefix for UUID-named run folders (the upload transaction id, which is also
   the `RunId`) and schedules any folder that no surviving run claims. It only ever *records*;
   deletion stays on the single audited path. Swept marker rows are now stamped `CompletedAt`
   instead of being deleted, which is how the pass remembers folders it has already handled —
   without it, the job would re-list every historical run folder every night.

Prefix deletion is far riskier than the per-key deletion the sweep did before, so it is fenced in:

- a recorded location must be under the lab prefix **and** contain the run id as a path segment,
  making it structurally impossible for a delete to escape one run's folder — a custom `outdir`
  pointing elsewhere is skipped and logged rather than deleted;
- a prefix is skipped while any surviving run still points at it (HealthOmics retries can reuse an
  `outdir`, and without this an expired run could delete a live run's results);
- reconciliation ignores folders touched within `ORPHAN_SCAN_MIN_AGE_DAYS` (default 30), because
  uploads land in the run folder *before* the run record exists — otherwise a user who uploaded
  files but has not launched their run yet would have their data scheduled for deletion. Such
  folders are deliberately left unrecorded so they are re-evaluated once they age;
- only `results/` and root-level `samplesheet*.csv` are ever touched. Input files stay with the
  FILE#-row sweep, which owns their reuse and Permanent-tag rules, and any key already tracked by
  a FILE# row is excluded;
- output deletes draw on the existing `MAX_DELETES_PER_LAB_SWEEP` budget, and
  `MAX_ORPHAN_FOLDERS_PER_LAB_SWEEP` (default 500) drains a large backlog over several nights
  rather than in one long invocation;
- `DRY_RUN` is honoured throughout — it lists and logs without deleting or recording.

## Testing*

- `pnpm jest` in `packages/back-end`: **125 suites / 1041 tests passing**, including the infra
  synth and `stack-resource-budget` guardrail tests that cover the IAM changes.
- 34 new unit tests across three suites:
  - `process-laboratory-run-stream.lambda.test.ts` — records the prefix and sample sheet, records
    alongside per-file unlinking, percent-decodes keys containing spaces, and refuses a custom
    `outdir` or a different bucket.
  - `process-expired-laboratory-data.lambda.test.ts` — deletes the whole `results/` subtree
    including the placeholder object, paginates the listing, skips a prefix a surviving run still
    uses, keeps the marker when the delete budget truncates a prefix, and does not re-sweep a
    completed marker. A separate suite covers reconciliation: recording an orphan, never
    scheduling input files, skipping live runs / already-known folders / recently-touched folders,
    honouring both new caps, and leaving a `results/` prefix alone when a tracked file row sits
    under it.
  - `s3-uri-utils.test.ts` — URI parsing and the run-folder containment check.
- `eslint` and `prettier --check` clean; `tsc --noEmit` introduces no new errors (the repo has
  pre-existing failures unrelated to this change).

**Not covered:** the unit tests mock S3, so they prove the logic but not real bucket layouts. See
Impact for the recommended rollout.

## Impact

**Behavioural change — this deletes data that previously survived.** Expired runs now lose their
`results/` tree and sample sheet, and the reconciliation pass reaches back over the existing
backlog. That is the intended fix, but it is irreversible, to test it deploy with `DRY_RUN=true`
for one cycle first and review the new EMF metrics under `EasyGenomics/DataRetention`:
`OrphanFoldersScanned`, `OrphanFoldersRecorded`, `OutputObjectsDeleted`, `SampleSheetsDeleted` and
`OutputPrefixesSkippedInUse`, alongside the `[dry-run] Would …` log lines. That gives a concrete
preview of the backlog size and what would be removed before anything is.

**IAM:** the sweep gains `s3:ListBucket` (it could already delete objects but not enumerate them)
and `dynamodb:Query` on `laboratory-run-table` for the live-run check; the stream subscriber gains
`dynamodb:PutItem` on the tagging table for the new row type.

**Runtime cost:** the nightly job now issues one delimiter listing per lab prefix plus one query,
and inspects only previously-unseen run folders. The `CompletedAt` design keeps this flat in
steady state instead of growing with run history.

**Storage:** `RUNOUTPUT#` rows are retained rather than deleted — roughly 150 bytes per run that
has ever existed. The tagging table has no TTL attribute configured, so they accumulate. Worth
revisiting if it ever becomes material.

**New configuration** (documented in the lambda's header comment, consistent with how the existing
`DRY_RUN` and `MAX_DELETES_PER_LAB_SWEEP` vars are documented): `ORPHAN_SCAN_MIN_AGE_DAYS`
(default 30) and `MAX_ORPHAN_FOLDERS_PER_LAB_SWEEP` (default 500).

## Additional Information

The gap was an integration oversight rather than a regression — output deletion was never
implemented, and git history confirms it was not present earlier and later removed.

Reviewers may want to focus on `reconcileOrphanedRunFolders`, since it is the only code path that
deletes by prefix based on inference rather than a recorded location.

## Checklist*


diegomayorga-dept:
# PR Review: fix: delete run outputs and sample sheets when a run retention expires (#947)

## Summary
- **3 blockers, 9 major, 7 minor**
- Top concerns: the live-run safety fence reads an **unpaginated** DynamoDB query, so it silently fails open and can delete a live run's outputs; the go-forward marker path has **no Permanent-tag/FILE# exclusion**; `DRY_RUN` is hardcoded `false` in CDK, so the backlog reconciliation goes destructive on the first 04:00 schedule after deploy.
- Three independent reviewers (correctness, security, breaking-changes) converged on the pagination blocker without seeing each other's output.

---

## Correctness (5)

### blocker — packages/back-end/src/app/controllers/easy-genomics/data-collections/process-expired-laboratory-data.lambda.ts:431
Both live-run fences (`liveRunIds` at 431, `livePrefixes` via `loadLiveRunOutputPrefixes` at ~607) are built from `laboratoryRunService.queryByLaboratoryId`, which issues a **single** `QueryCommand` and never follows `LastEvaluatedKey` (`laboratory-run-service.ts:137-162`). DynamoDB caps a page at 1MB; run rows carry `InputFileKeys`/`Settings`, so a few hundred runs exceeds it. Runs past page 1 are silently absent → `reconcileOrphanedRunFolders` classifies a **surviving** run's folder as orphaned, and `isPrefixStillInUse` then returns false for that same prefix, so the live run's entire `results/` tree and sample sheet are irreversibly deleted. Note the sibling loads in the same function (`listAllFileRowsForLab`, `listExpiredRunOutputsForLab`) *do* paginate — this is the one unpaginated input, and it is the one driving destructive decisions. This directly invalidates the PR body's claim that "a prefix is skipped while any surviving run still points at it."
**Suggestion:** Add a paginated `listAllRunsForLab` (loop on `LastEvaluatedKey`) and use it in both call-sites. Fail closed: if the query returns a `LastEvaluatedKey`, abort reconciliation and the output sweep for that lab and emit an error metric rather than proceeding on a partial live set.

### blocker — packages/back-end/src/app/controllers/easy-genomics/data-collections/process-expired-laboratory-data.lambda.ts:348
The marker path deletes every key under `row.OutputPrefix` with **no FILE#-row / Permanent-tag exclusion**. `add-tags-to-files.lambda.ts` accepts arbitrary object keys and the tagging service only checks `assertKeyUnderLabPrefix`, so a user can mark a file under `results/` Permanent from the File Manager. When that run TTLs out, `listAllKeysUnderPrefix` + `deleteKeysInBatches` remove the Permanent-tagged object anyway, defeating the protection `isPermanent` enforces at line 217 — and the deleted objects' FILE# rows are left dangling. The *reconciliation* path has a `resultsAreTracked` guard for exactly this; the go-forward stream-marker path — the main path of this PR — has none.
**Suggestion:** Load the lab's FILE# rows once per lab (already loaded for reconciliation) and filter out any listed key carrying the lab's permanent tag id, or skip the prefix entirely as reconciliation does; delete the corresponding FILE# rows for keys that are removed.

### major — packages/back-end/src/app/utils/s3-uri-utils.ts:18
`parseS3Uri` uses `new URL` on a raw `s3://bucket/key`, so a literal `#` in the key starts a fragment and the key is **silently truncated**; a lone `%` makes `decodeURIComponent` throw (caught → null). Both characters are explicitly permitted by `buildSampleSheetFileName` (front-end `sample-sheet-utils.ts` allows `[a-zA-Z0-9._:!@#$%^()-]`) and the URL is stored unencoded. A run named "Run #5" → `samplesheet-Run-#5.csv` is recorded as `.../samplesheet-Run-`. `DeleteObjects` reports success for non-existent keys, so `removed === keys.length` holds, the marker is stamped `CompletedAt`, and reconciliation skips the folder forever via `knownRunIds` — the real sample sheet is orphaned permanently. The `decodeURIComponent` also corrupts any key genuinely containing `%20`.
**Suggestion:** Parse by string split, not `URL`: `const rest = value.slice(5); const i = rest.indexOf('/')`. Drop the percent-decoding (the stored URL is already the raw key), or restrict it to values from a genuinely encoded source.

### major — packages/back-end/src/app/controllers/easy-genomics/laboratory/run/process-laboratory-run-stream.lambda.ts:141
The only fence on the recorded `OutputPrefix` is `isWithinRunFolder` — under the lab prefix and containing `/{runId}/` as a segment. `outdir` is a **user-editable** workflow parameter (defaulted in `run-upload-sample-sheet.ts` / `EGRunFormUploadData.vue`, carried as `OutputS3Url`), so a user setting `outdir` to the run folder *root* passes the guard. The sweep then prefix-deletes the whole run folder — which is where that run's **input FASTQ files** live — bypassing the FILE#-row sharing and Permanent-tag rules. Nothing downstream re-restricts to `results/`.
**Suggestion:** Reject a normalized prefix equal to the run-folder root, or only record prefixes ending in `results/`; anything else falls through to the FILE#-row sweep.

### major — packages/back-end/src/app/controllers/easy-genomics/data-collections/process-expired-laboratory-data.lambda.ts:335
`isPrefixStillInUse` is gated on `row.OutputPrefix` existing, and `livePrefixes` is built only from `OutputS3Url`. A marker row carrying **only** `SampleSheetKeys` (reconciliation writes these when `resultsAreTracked`; legacy runs have no `OutputS3Url`) is deleted with no live-run check at all. The retry flow (`prefillFromFailedRun`) copies the failed run's `SampleSheetS3Url` verbatim into a new run with a new RunId, so a live retry's sheet lives in the expired run's folder and is deleted underneath it.
**Suggestion:** Collect surviving runs' `SampleSheetS3Url` (and `InputS3Url` for legacy runs) into the live set and filter each `SampleSheetKeys` entry against it, counting skips in `skippedInUse`.

## Security (2)

### blocker — packages/back-end/src/app/controllers/easy-genomics/data-collections/process-expired-laboratory-data.lambda.ts:431
Same unpaginated-query finding as above, reached independently — see Correctness. Security framing: `isPrefixStillInUse` **fails open** on truncation, and `results/` objects are excluded from the `trackedKeys` FILE# exclusion, so the pass writes a `RUNOUTPUT#` row for live data and then deletes it in the same invocation.

### major — packages/back-end/src/app/controllers/easy-genomics/data-collections/process-expired-laboratory-data.lambda.ts:322
The delete path re-validates a `RUNOUTPUT#` row only with `assertKeyUnderLabPrefix` (lab root), **never** with `isWithinRunFolder` — that check runs only at *write* time in the stream lambda. So "a delete can never escape one run's folder" is enforced by the writer, not the deleter: a row with `OutputPrefix = org-1/lab-1/` would wipe the entire laboratory prefix, Permanent-tagged files included. The reviewer traced both current writers and confirmed neither can produce such a row today — so it is **not exploitable as written**, but it is the fence the PR body calls structural, and it is one future writer or one bad-data row away from a lab-wide delete.
**Suggestion:** In `sweepLaboratoryRunOutputs`, apply `isWithinRunFolder(row.OutputPrefix, laboratory, row.RunId)` (and per sample-sheet key) at delete time, skipping the row on failure.

## Breaking changes (1)

### blocker — packages/back-end/src/infra/stacks/easy-genomics-nested-stack.ts:416
`DRY_RUN` is hardcoded `'false'` for `process-expired-laboratory-data`, so both the new reconciliation pass and the prefix-delete sweep go **live and destructive on the first 04:00 schedule after deploy**, in the same invocation (reconcile runs immediately before `sweepLaboratoryRunOutputs`). This deletes pre-existing run outputs and sample sheets that previously survived — irreversible absent bucket versioning — with no separate flag, allowlist or per-lab gate to stage it. The PR body *recommends* a `DRY_RUN=true` cycle first, but nothing in the code enforces it; it requires an out-of-band manual stack edit.
**Suggestion:** Gate the new reconciliation/output-delete path behind its own env flag defaulting to off (or record-only) so the first deploy audits, and set `MAX_ORPHAN_FOLDERS_PER_LAB_SWEEP` explicitly in the stack for the first rollout rather than relying on the code default.

## Conventions (2)

### major — packages/back-end/src/app/controllers/easy-genomics/data-collections/process-expired-laboratory-data.lambda.ts:411
The handler grows to **703 lines** — the largest controller in the repo — with ~370 lines of new domain logic (`sweepLaboratoryRunOutputs`, `reconcileOrphanedRunFolders`, bucket-walking helpers) in a `*.lambda.ts`. EG's rule is thin controllers (≤~50 lines) with policy in a domain service; orphan discovery, live-prefix arbitration, delete budgeting and marker lifecycle are exactly that. Sibling `process-*` handlers are 24–180 lines and delegate.
**Suggestion:** Move the sweep/reconciliation logic into a `laboratory-data-retention-service.ts` (or onto `LaboratoryDataTaggingService`, which already owns the `RUNOUTPUT#` accessors), leaving the handler to iterate labs and emit metrics.

### minor — packages/back-end/src/app/controllers/easy-genomics/data-collections/process-expired-laboratory-data.lambda.ts:512
`lastPathSegment`, `isRunFolderName`, `isSampleSheetKey` are `export`ed from a `*.lambda.ts` with **no importer anywhere** (not even the new test file), and `lastPathSegment` duplicates the private one in `data-collection-service.ts:12`. The PR creates `s3-uri-utils.ts` for exactly this shape of helper.
**Suggestion:** Move them (and their regex constants) into `s3-uri-utils.ts`, or drop the `export` keyword.

## Tests (4)

### major — packages/back-end/src/app/controllers/easy-genomics/data-collections/process-expired-laboratory-data.lambda.ts:644
`deleteKeysInBatches` has **no partial-failure test** — every test mocks `deleteObjects` as `{ Errors: [] }`. The `removed = chunk.length - Errors.length` arithmetic feeds the safety invariant that the marker is stamped `CompletedAt` only when `removed === keys.length` (line 371). A regression ignoring `response.Errors` would stamp complete while objects survive, orphaning them permanently (the row is never re-swept; reconciliation skips completed markers). The >1000-key chunking loop is likewise untested.
**Suggestion:** Test `{ Errors: [{ Key, Code: 'AccessDenied' }] }` → assert `markExpiredRunOutputCompleted` NOT called; and a >1000-key list → assert two `deleteObjects` calls.

### major — packages/back-end/src/app/services/easy-genomics/laboratory-data-tagging-service.ts:1250
All three new public methods have **zero coverage** — the only references under `test/` are `jest.mock` stubs, so the real implementations never run, despite a ~1300-line existing suite with a `describe` per method. `listExpiredRunOutputsForLab` in particular has untested `LastEvaluatedKey` pagination (a dropped page = expired outputs silently never deleted) and defensive per-field type filtering.
**Suggestion:** Add a `describe` block following the suite's aws-sdk-client-mock pattern: assert the `begins_with(Sk, 'RUNOUTPUT#')` key condition, two-page pagination, malformed-row dropping, and that `markExpiredRunOutputCompleted` issues a SET rather than a delete.

### minor — packages/back-end/src/app/controllers/easy-genomics/data-collections/process-expired-laboratory-data.lambda.ts:31
Env guards for a destructive job are only happy-path tested. The documented kill-switch `MAX_DELETES_PER_LAB_SWEEP=0` has no test, nor does the malformed/negative fallback to 25,000 — a typo'd value silently falling back to the 25k default is the exact failure mode these fences exist to prevent.

### minor — packages/back-end/src/infra/stacks/easy-genomics-nested-stack.ts:73
None of the infra changes are covered, although the stack test file is assertion-based with an established per-change pattern. Untested: the new `s3:ListBucket`, `dynamodb:Query` on the run table, `dynamodb:PutItem` for the stream lambda, plus the new `classificationDlq` and its `maxReceiveCount: 3` wiring. A lost `ListBucket` grant turns the delete path into a **silent no-op** (errors are caught and counted, not thrown).

## Simplification (4)

### minor — packages/back-end/src/app/utils/s3-uri-utils.ts:10
`parseS3Uri` is a **fourth** copy — identical bodies already exist inline in `request-top-level-bucket-objects.lambda.ts:26`, `request-search-bucket-objects.lambda.ts:26` and `request-folder-download-job.lambda.ts:37`; `normalizeS3Prefix` duplicates `normalizePrefix` in the last of those. Extracting is right, but the originals were left, so the net effect is one more copy, not one fewer — and they can now drift (the new one percent-decodes, the old three don't).

### minor — packages/back-end/src/app/controllers/easy-genomics/data-collections/process-expired-laboratory-data.lambda.ts:621
`listAllKeysUnderPrefix` is `listObjectsWithMetadataUnderPrefix` with `.map(o => o.Key)`; `listChildPrefixes` is a third copy of the same loop differing only by `Delimiter`. The same loop is hand-rolled in five other handlers. `s3-service.ts` — which this PR already extends — is the established home.

### minor — packages/back-end/src/app/controllers/easy-genomics/data-collections/process-expired-laboratory-data.lambda.ts:40
The new generic `parseNumericEnv` sits eight lines below `parseMaxDeletesPerLabSweep`, whose body is character-for-character the same logic inlined. Collapse the old one into the new.

### minor — packages/back-end/src/app/controllers/easy-genomics/data-collections/process-expired-laboratory-data.lambda.ts:53
`ORPHAN_SCAN_MIN_AGE_DAYS` and `MAX_ORPHAN_FOLDERS_PER_LAB_SWEEP` are wrapped in zero-arg functions with one call site each, and **neither is set anywhere in CDK or `config/`**. The 30-day quiet period is a correctness invariant of the reconciliation pass, not an operational dial — making it runtime-tunable invites it being set to something unsafe.

## PR hygiene (2)

### major — PR
No ticket reference. The title uses a conventional-commit prefix instead of the repo's `EG-`/`EGV-` convention, and the body still carries the unfilled placeholder `<!-- Add the EG- ticket reference here if there is one. -->`.

### major — PR
~655 net hand-written source lines across 6 source files vs the repo's ~400-line budget, with no split proposal or approval note. The PR also bundles two independently shippable concerns: the forward-looking `RUNOUTPUT#` recording path, and the historical reconciliation/orphan-scan pass — which is also the riskier half.

diegomayorga-dept (approved):

